PR #88, opened 2023-09-08
Noelle & Tommy Adopt Don't Shop

juliet-e:
We're creating this to leave comments on your work!


app/controllers/applicants_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+    end
+  end
+
+  def new
+
+  end
+
+  def create
+    applicant = Applicant.new(applicant_params)
+    if applicant.save
```
juliet-e, line 23: Nice error handling!

app/controllers/applicants_controller.rb
```diff
@@ -0,0 +1,35 @@
...
+  
+  def show
+    @applicant = Applicant.find(params[:id])
+    @pets = Pet.all
+    
+    if params[:name].present? 
+      @results = @pets.search(params[:name]) 
+      @search = params[:name]
+      # redirect_to "/applicants/#{@applicant.id}"
+    end
```
juliet-e, line 14: It looks like your button to Add a Pet to an application redirects back to this action using this path: `/applicants/:applicant_id` and then with query parameters for a pet ID

That would mean you have to create your join record in this controller method, which you probably don't want to do (SRP). Another option for that button could be to route it to a `applicant_pets_controller` and specifically do a `create` action there. In that `create` method, you could do something like `PetApplicant.create` with the applicant_id and pet_id. The path for that button could be `/applicants/:applicant_id/pets/:pet_id` or you could also decide to send the pet ID as a hidden field if you didn't want it in the path

app/controllers/pet_applicants_controller.rb
```diff
@@ -0,0 +1,9 @@
+class PetApplicantsController < ApplicationController
+  def index
+  end
+
+  def create
+    PetApplicant.create!(pet: Pet.find(params[:pet]), applicant: Applicant.find(params[:id]))
```
juliet-e, line 6: Boom! You already have this controller made! So in order to get your application to arrive at this action, you would need to update your button on the applicant show page to be:
`<%= button_to "Adopt this Pet", "/applicants/#{@applicant.id}/pets/#{result.id}", method: :post %>`
Note that you don't need to add any Turbo business because Turbo actually doesn't affect buttons! Just links and forms. 

For this to work, you'd also need to update your routes. Seem my comment in your routes file

PR #90, opened 2023-09-09
Lane & John Adopt Don't Shop Eval

dcoleman21:
Overall nice job you two. I have left some comments for you to review.


app/controllers/admin/applicants_controller.rb
```diff
@@ -0,0 +1,37 @@
+class Admin::ApplicantsController < ApplicationController
+  def index
+    @applicants = Applicant.where(application_status: "Pending")
```
dcoleman21, line 3: So here you are using the `.where` AR method. I think you would probably want to move this to the `applicants` model and make a method that you can call here in your controller.

app/controllers/admin/applicants_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+  
+  def show
+    @applicant = Applicant.find(params[:id])
+    @applicant_pets = @applicant.pets
```
dcoleman21, line 8: So in the spirit of keeping our controllers lean, I would remove this and then on line 12 in your `applicants/show.html.erb` I would just call `@applicant.pets.each do`.

app/controllers/admin/applicants_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def approve_pet
```
dcoleman21, line 11: I think both your `approve_pet` and `reject_pet` methods could potentially live in a `update` action with some control flow `if/else`. However, I can also see how this could help with readability for other devs.

app/controllers/admin/shelters_controller.rb
```diff
@@ -0,0 +1,9 @@
+class Admin::SheltersController < ApplicationController
+  def index
+    @shelters = Shelter.find_by_sql("SELECT * FROM shelters ORDER BY name DESC;") #order(name: :desc)
```
dcoleman21, line 3: All of the AR and SQL that is taking place here should live at the model level. Then you will just call those methods here in the controller.

app/controllers/applicants_controller.rb
```diff
@@ -0,0 +1,43 @@
+class ApplicantsController < ApplicationController
+  def show
+    @applicant = Applicant.find(params[:id])
+    @applicant_pets = @applicant.pets
```
dcoleman21, line 4: You don't need to have this additional instance variable here. In your view you can call `@applicant.pets.each do`. This keeps your controller lean.

app/controllers/applicants_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    @applicants = Applicant.all
+  end
+
+  def new
+  end
+
+  def create
+    @applicant = Applicant.new(applicant_params)
+    if @applicant.save
+      redirect_to "/applicants/#{Applicant.last.id}"
```
dcoleman21, line 24: here I think you could probably do 
```ruby
redirect_to "/applicants/#{@applicant.id}"
```
where you are grabbing the `id` that is created with your `.new` on line 22.

app/controllers/applicants_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+    else
+      flash[:error] = 'Invalid data. Please fill out fields correctly.'
```
dcoleman21, line 26: Nice job with the flash message

app/models/pet.rb
```diff
@@ -2,6 +2,8 @@ class Pet < ApplicationRecord
   validates :name, presence: true
   validates :age, presence: true, numericality: true
   belongs_to :shelter
+  has_many :applicants_pets
```
dcoleman21, line 5: Nice job with your associations

app/views/admin/applicants/index.html.erb
```diff
@@ -0,0 +1,6 @@
+<h1>All Submitted Applications</h1>
+
+<% @applicants.each do |applicant| %>
+  <%= render partial: "applicants/applicant_info", locals: {object: applicant} %>
```
dcoleman21, line 4: Nice use of a partial here.

app/views/admin/applicants/show.html.erb
```diff
@@ -0,0 +1,27 @@
...
+<%= @applicant.street_address %><br/>
+<%= @applicant.city %>, <%= @applicant.state %> <%= @applicant.zip_code %></p>
+<p>Why I would make a good home:<br/>
+<%= @applicant.description %></p>
+<p>Application Status:<br/>
+<%= @applicant.application_status %></p>
+
+<h3>Pets being applied for:</h3>
+<ul>
+<% @applicant_pets.each do |pet| %>
```
dcoleman21, line 12: here is where you can remove the `@applicant_pets` instance variable from the controller and just call `@applicant.pets`.

spec/features/admin/applicants/index_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    @pet_1 = Pet.create!(adoptable: true, age: 1, breed: "sphynx", name: "Lucille Bald", shelter_id: @shelter1.id)
+    ApplicantsPet.create!(applicant: @bob, pet: @pet_1)
+  end
+
+  describe "As a visitor" do
+    describe "When I visit the admin applicants index page" do
+      it "displays all applicants who have applications submitted and their attributes" do
+        visit "/admin/applications"
+
+        expect(page).to have_link("Applicants Show page")
```
dcoleman21, line 16: Using `within` blocks is a great way to make your tests even more robust. If we think about being on an index page, we are getting all of our objects displayed. So by using `within` blocks we can really isolate where something is on the page and expect that section to not have things as well.

spec/features/admin/show_spec.rb
```diff
@@ -0,0 +1,97 @@
...
+      ApplicantsPet.create(applicant: @bob, pet: @pet_2)
+  end
+
+  # User Story 12 Test
+  it 'For every pet that the application is for, I see a button to approve the application ' do
+    visit "/admin/applications/#{@bob.id}"
+
+
+    
+    within("##{@pet_1.id}") do
```
dcoleman21, line 53: Great job using `within` blocks here. Keep working to implement them in all of your tests.

spec/models/applicant_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  describe "instance methods" do
+    before :each do
+      @applicant = Applicant.create!(name: "Bob", street_address: "1234 Bob's Street", city: "Fudgeville", state: "AK", zip_code: 27772, description: "", application_status: "In Progress")
+      @shelter = Shelter.create!(name: "Aurora shelter", city: "Aurora, CO", foster_program: false, rank: 9)
+      @pet_1 = @applicant.pets.create!(adoptable: true, age: 2, breed: "Dog", name: "Rex", shelter_id: @shelter.id)
+    end
+
+    it "#descript" do
```
dcoleman21, line 26: Nice job testing your model methods.

PR #92, opened 2023-09-09
Allan & Nick M Adopt Don't Shop Eval

dcoleman21:
Hey ya'll, I have left some comments on your code. I tried to address the questions that you had in your video. Please let me know if you have any more. Nice job.

### DIFF
--- a/app/controllers/admin_shelters_controller.rb
+++ b/app/controllers/admin_shelters_controller.rb
@@ -1,5 +1,6 @@
 class AdminSheltersController < ApplicationController
   def index
     @shelters = Shelter.find_by_sql("SELECT * FROM shelters ORDER BY name DESC")
+    # require 'pry';binding.pry
   end
 end
--- a/app/controllers/admin_shelters_controller.rb
+++ b/app/controllers/admin_shelters_controller.rb
@@ -1,0 +1,5 @@
+class AdminSheltersController < ApplicationController
+  def index
+    @shelters = Shelter.find_by_sql("SELECT * FROM shelters ORDER BY name DESC")
+  end
+end
--- a/app/controllers/admin_shelters_controller.rb
+++ b/app/controllers/admin_shelters_controller.rb
@@ -3,4 +3,21 @@ class AdminSheltersController < ApplicationController
     @shelters = Shelter.find_by_sql("SELECT * FROM shelters ORDER BY name DESC")
     # require 'pry';binding.pry
   end
+
+  def show
+    @application = Application.find(params[:app_id])
+  end
+
+  def update
+    application = Application.find(params[:app_id])
+    application.pet_applications.each do |pa|
+      if pa.id == params[:pa_id].to_i
+        pa.update({
+          pet_app_status: [params[:new_status]]
+        })
+        pa.save
+      end
+    end
+      redirect_to "/admin/shelters/#{application.id}"
+  end
 end
--- a/app/controllers/admin_shelters_controller.rb
+++ b/app/controllers/admin_shelters_controller.rb
@@ -1,7 +1,6 @@
 class AdminSheltersController < ApplicationController
   def index
     @shelters = Shelter.find_by_sql("SELECT * FROM shelters ORDER BY name DESC")
-    # require 'pry';binding.pry
   end
 
   def show
@@ -15,9 +14,11 @@ class AdminSheltersController < ApplicationController
         pa.update({
           pet_app_status: [params[:new_status]]
         })
+        application.update(status: params[:new_status])
         pa.save
       end
     end
-      redirect_to "/admin/shelters/#{application.id}"
+
+    redirect_to "/admin/shelters/#{application.id}"
   end
 end
--- a/app/controllers/admin_shelters_controller.rb
+++ b/app/controllers/admin_shelters_controller.rb
@@ -9,14 +9,20 @@ class AdminSheltersController < ApplicationController
 
   def update
     application = Application.find(params[:app_id])
-    application.pet_applications.each do |pa|
-      if pa.id == params[:pa_id].to_i
-        pa.update({
-          pet_app_status: [params[:new_status]]
-        })
-        application.update(status: params[:new_status])
-        pa.save
+    if params[:pa_id].present?
+      application.pet_applications.each do |pa|
+        if pa.id == params[:pa_id].to_i
+          pa.update({
+            pet_app_status: params[:new_status]
+          })
+          pa.save
+        end
       end
+    elsif params[:app_status].present?
+      application.update({
+        status: params[:app_status]
+      })
+      application.save
     end
 
     redirect_to "/admin/shelters/#{application.id}"
--- a/app/controllers/applications_controller.rb
+++ b/app/controllers/applications_controller.rb
@@ -7,7 +7,7 @@ class ApplicationsController < ApplicationController
     @application = Application.find(params[:app_id])
 
     if params[:search].present?
-      @searched_pets = Pet.where("lower(name) LIKE ?", "%#{params[:search].downcase}%") 
+      @searched_pets = Pet.search_pets(params[:search])
     end
   end
 
@@ -29,10 +29,11 @@ class ApplicationsController < ApplicationController
 
   def update
     application = Application.find(params[:app_id])
-    application.update({
+    
+    application.update(
       description: params[:reason],
-      status: [params[:app_status]]
-    })
+      status: params[:app_status]
+    )
     application.save
     redirect_to "/applications/#{application.id}"
   end
--- a/app/controllers/applications_controller.rb
+++ b/app/controllers/applications_controller.rb
@@ -24,6 +24,12 @@ class ApplicationsController < ApplicationController
     end
   end
 
+  def add_pet
+    application = Application.find(params[:app_id])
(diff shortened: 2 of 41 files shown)


app/controllers/admin_shelters_controller.rb
```diff
@@ -0,0 +1,30 @@
+class AdminSheltersController < ApplicationController
+  def index
+    @shelters = Shelter.find_by_sql("SELECT * FROM shelters ORDER BY name DESC")
```
dcoleman21, line 3: This would be considered an MVC infraction because you are doing SQL logic here in the controller. We would want to move this logic to the shelter model in some method. Then we would call that method here in the controller.

app/controllers/admin_shelters_controller.rb
```diff
@@ -0,0 +1,30 @@
+class AdminSheltersController < ApplicationController
```
dcoleman21, line 1: You would want to create a directory called `admin` and then within that directory have a `shelters` controller and an `applications` controller. 
```ruby 
app/controllers/admin/shelters_controller.rb
app/controllers/admin/applications_controller.rb
``` 
Then your class would read
```ruby
class Admin::SheltersController < ApplicationController
  def index
     @shelters = Shelter.list_alphabetically
     @pending_applications = Shelter.pending_applications
  end
end
```
in here is where you will call your model method to alphabetize all shelters. As well as possibly call a method that you create to get all pending applications

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+  
+  def create
+    @application = Application.new(application_params)
+    @application.status = "In Progress"
+    
+    if @application.save
+      redirect_to "/applications/#{@application.id}"
+    else
+      redirect_to "/applications/new"
```
dcoleman21, line 22: Nice use of flash messaging

app/models/application.rb
```diff
@@ -0,0 +1,16 @@
...
+  has_many :pets, through: :pet_applications
+
+  validates :name, presence: true
+  validates :address, presence: true
+  validates :city, presence: true
+  validates :state, presence: true
+  validates :zip, presence: true
+  validates :description, presence: true
+
+  def add_pet(pet_id)
```
dcoleman21, line 12: I can see what you were trying to do here, add a pet to an application. I think that can be done with the join `pet_application`. So you might have wanted to create a `pet_application_controller` with a `create` action for that association.

app/models/pet.rb
```diff
@@ -10,4 +12,10 @@ def shelter_name
   def self.adoptable
     where(adoptable: true)
   end
+
+  def self.search_pets(search_params)
+    if search_params.present?
```
dcoleman21, line 17: Nice job here.

app/models/pet_application.rb
```diff
@@ -0,0 +1,4 @@
+class PetApplication < ApplicationRecord
+  belongs_to :pet
+  belongs_to :application
```
dcoleman21, line 3: Great job creating the association. You never utilized this by creating a join controller.

app/views/admin_shelters/index.html.erb
```diff
@@ -0,0 +1,16 @@
+<% @shelters.each do |shelter| %>
+  <h3><%= shelter.name %></h3>
+<% end %>
+
+
+<h2> Shelters with Pending Applications </h2>
+
+<% @shelters.each do |shelter| %>
+  <% shelter.pets.each do |pet| %>
```
dcoleman21, line 9: See my comment [here](https://github.com/turingschool-examples/adopt-dont-shop-7/pull/92/files#r1320598681) about this.

app/views/admin_shelters/show.html.erb
```diff
@@ -0,0 +1,21 @@
...
+<%= @application.status %>
+
+  <% if !@application.pets.empty? && @application.status == "Pending" %>
+    <%= button_to "Approve Application", "/admin/shelters/#{@application.id}?app_status=Accepted", method: :patch%>
+  <% end %>
+  <% @application.pets.each do |pet| %>
+    <% pet.pet_applications.each do |pa| %>
+      <% if pa.application_id == @application.id && pa.pet_id == pet.id %>
+        <% if pa.pet_app_status == nil && @application.status == "Pending" %>
+          <%= button_to "Approve #{pet.name}", "/admin/shelters/#{@application.id}?new_status=Approved&pa_id=#{pa.id}", method: :patch%>
```
dcoleman21, line 12: nice job using query params

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,41 @@
...
+<h3><%= @application.city %></h3>
+<h3><%= @application.state %></h3>
+<h3><%= @application.zip %></h3>
+<h3><%= @application.description %></h3>
+<h3><%= @application.status %></h3>
+
+<%if !@application.pets.empty? %>
+  <% @application.pets.each do |pet| %>
+    <h5><%= link_to "#{pet.name}", "/pets/#{pet.id}", method: :get%></h5>
+    <%= render partial: "applications/pet_attributes", locals: { pet: pet } %>
```
dcoleman21, line 12: Nice job using partials

app/views/pets/show.html.erb
```diff
@@ -1,8 +1,3 @@
-<h1><%= @pet.name %></h1>
-<h3><%= @pet.age %> years old</h3>
-<h3><%= @pet.breed %></h3>
-<h3><%= @pet.adoptable %></h3>
-<h3><%= @pet.shelter_name %></h3>
-
+<%= render partial: "pets/pet_list", locals: { pet: @pet } %>
```
dcoleman21, line 1: If this partial lives in the `pets` directory, you can just call the name of the partial without the `pets` prefix.

config/routes.rb
```diff
@@ -1,6 +1,13 @@
 Rails.application.routes.draw do
+  root 'welcome#index'
   get "/", to: "application#welcome"
 
+  post "/applications", to: "applications#create"
+  get "/applications/new", to: "applications#new"
+  get "/applications/:app_id", to: "applications#show", as: "application"
+  patch "/applications/:app_id", to: "applications#update"
+  patch "/applications/add_pet/:app_id", to: "applications#add_pet"
```
dcoleman21, line 9: Here is where I think you probably could have created a `pet_applications` controller and done something like
```
patch "/admin/applications/:id", to: "pet_applications#update"
```
which will keep your routes RESTful and lean on join table you created. You would also have to change the url on the form.

config/routes.rb
```diff
@@ -36,4 +43,8 @@
   get "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinary_offices#veterinarians"
   get "/veterinary_offices/:veterinary_office_id/veterinarians/new", to: "veterinarians#new"
   post "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinarians#create"
+  
+  get "/admin/shelters", to: "admin_shelters#index"
```
dcoleman21, line 47: So we wouldn't want this to point to a `admin_shelters` controller but rather a directory called admin, which you are doing with `"/admin"` that then points to a file called `shelters` with an `index` action. 
```
get "/admin/shelters", to: "admin/shelters#index"
```

db/migrate/20230831205941_create_pet_applications.rb
```diff
@@ -0,0 +1,11 @@
+class CreatePetApplications < ActiveRecord::Migration[7.0]
+  def change
+    create_table :pet_applications do |t|
+      t.references :pet, null: false, foreign_key: true
+      t.references :application, null: false, foreign_key: true
+      t.string :pet_app_status
```
dcoleman21, line 6: I don't think that you would need to add this column here. Your applications table has a status on it. So you would just need to use ActiveRecord to reach through the `pet_applications` table 
to get to the status attribute on your applications table. You can do this because 
```
a pet has_many :pet_applications
and has_many :applications, through: :pet_applications
```

spec/features/admin_shelters/index_spec.rb
```diff
@@ -0,0 +1,31 @@
+require "rails_helper"
+
+RSpec.describe "the pets index" do
+  it "Lists shelters in decending order" do
+    shelter_1 = Shelter.create(name: "Aurora shelter", city: "Aurora, CO", foster_program: false, rank: 9)
+    shelter_2 = Shelter.create(name: "Ford shelter", city: "Aurora, CO", foster_program: false, rank: 9)
+    shelter_3 = Shelter.create(name: "Zeta shelter", city: "Aurora, CO", foster_program: false, rank: 9)
+
+    visit '/admin/shelters'
+
```
dcoleman21, line 10: I would wrap these up in `within` blocks to better test.

spec/models/application_spec.rb
```diff
@@ -0,0 +1,17 @@
+require "rails_helper"
+
+RSpec.describe Application, type: :model do
+  describe "relationships" do
+    it { should have_many :pet_applications }
+  it { should have_many(:pets).through(:pet_applications) }
+  end
+
+  it "#add_pet" do
+    shelter = Shelter.create(name: "Aurora shelter", city: "Aurora, CO", foster_program: false, rank: 9)
```
dcoleman21, line 10: Nice job testing

PR #93, opened 2023-09-09
Kaylee & Nick S Adopt Don't Shop Eval 

dcoleman21:
Overall nice job you two. I left some comments for you to review. I am not entirely sure what was going on with the id holding two different states without getting into the code deeper. I do wonder if it has something to do with the association of an applicant and pet because you had to chain `applicant.id.pet` (something like this) and so the status lives on the pet and not the applicant so I wonder if that dependency has something to do with it. Not sure :/


app/controllers/admin_controller.rb
```diff
@@ -0,0 +1,25 @@
+class AdminController < ActionController::Base
```
dcoleman21, line 1: So when it comes to making admin type directories we will want to do it in such a way that we can both stay RESTful and can separate functionality from other controllers. So for the admin one, you will probably want to create a directory called `admin` and inside of that directory have a file called `shelters_controller.rb` and possibly `applications_controller.rb`. Now you can remain RESTful with CRUD actions. The class would look like:
```ruby
class Admin::SheltersController < ApplicationController

end
```
Now you can have an index action that is essentially doing what your `shelters_index` action is doing.

app/controllers/admin_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+  end
+
+  def show
+    @application = PetsApplication.find(params[:id])
+    @applicant = Applicant.find(@application.applicant_id)
+    @pets = Pet.pets_with_app_status_by_sql(@application)
+    @status = PetsApplication.check_app_status(@applicant)
+  end
+
+  def approve_reject
```
dcoleman21, line 17: I am wondering if this could live on your `update` action in your `pet_applications` controller??

app/controllers/pets_applications_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+    @application = PetsApplication.find(params[:id])
+    @applicant = Applicant.retrieve_applicant(params[:id])
+    @pets = @applicant.pets
+    @status = PetsApplication.check_app_status(@applicant)
+  end
+
+  def create
+    applicant = Applicant.new(applicant_params)
+    if applicant.save
+      application = PetsApplication.create!(applicant: applicant)
```
dcoleman21, line 16: In here I think you could create a new association by passing `PetApplication.create!` both a `pet_id` and an `application_id`

app/controllers/pets_applications_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+    application = PetsApplication.find(params[:id])
+    application.pet_id.nil? ? update_empty_pet_app(application) : create_new_or_update_app(application)
+    applicant = Applicant.find(application.applicant_id)
+    applicant.update(description: params[:description])
+    redirect_to "/pets_applications/#{application.id}"
+  end
+
+  private 
+
+  def applicant_params
```
dcoleman21, line 34: Nice job using strong params and creating additional private methods

app/models/pet.rb
```diff
@@ -10,4 +12,19 @@ def shelter_name
   def self.adoptable
     where(adoptable: true)
   end
+
+  def self.pets_with_app_status_by_sql(application)
+    find_by_sql("SELECT pets.name, pets.id, pets_applications.status FROM pets
```
dcoleman21, line 17: great job using raw SQL

app/models/shelter.rb
```diff
@@ -31,4 +31,12 @@ def alphabetical_pets
   def shelter_pets_filtered_by_age(age_filter)
     adoptable_pets.where("age >= ?", age_filter)
   end
+
+  def self.order_by_name_alphabetical
+    find_by_sql("SELECT * FROM shelters ORDER BY name DESC;")
+  end
+
+  def self.with_pending_applications
+    Shelter.joins(pets: :pets_applications).distinct.pluck(:name).sort
```
dcoleman21, line 40: you could probably also joins on `(pets: :applications)`

app/views/admin/_button.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= form_with url: url, method: method, data: {turbo: false} do |f| %>
```
dcoleman21, line 1: Nice partial

app/views/pets_applications/show.html.erb
```diff
@@ -0,0 +1,43 @@
...
+  state: @applicant.state,
+  zip_code: @applicant.zip_code,
+  description: @applicant.description
+} %>
+</section>
+
+<section class="search_adopt">
+<% if @application.status == "In Progress" %>
+<h2>Add a Pet to this Application</h2>
+<%= form_with url: "/pets_applications/#{@application.id}", method: :get, data: { turbo: false } do |form| %>
```
dcoleman21, line 15: I am wondering if this route could be more RESTful if we did
```
"/applications/:id/pets"
```
That way we are associating an application with a pet by passing in both ids

config/routes.rb
```diff
@@ -36,4 +36,15 @@
   get "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinary_offices#veterinarians"
   get "/veterinary_offices/:veterinary_office_id/veterinarians/new", to: "veterinarians#new"
   post "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinarians#create"
-end
+
+  get "/pets_applications/new", to: "pets_applications#new"
```
dcoleman21, line 40: This section is where I might consider creating routes like
```ruby
"/applications"
"/applications/:id" (show/patch)
"applications/new"
"applications/:id/pets", to: "pet_applications#create"
```

spec/features/admin/index_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    applicant_1 = Applicant.create(name: "George", street_address: "1421 W Mockingbird Ln", city: "Boulder", state: "Texas", zip_code: "23452", description: "I love animals!")
+
+    app_1 = PetsApplication.create!(applicant: applicant_1, pet: pet_1 )
+    app_2 = PetsApplication.create!(applicant: applicant_1, pet: pet_2 )
+    app_3 = PetsApplication.create!(applicant: applicant_1, pet: pet_3 )
+    app_4 = PetsApplication.create!(applicant: applicant_1, pet: pet_4 )
+
+    visit "/admin/pets_applications"
+
```
dcoleman21, line 22: I would wrap this in a `within` block to really get robust

spec/features/admin/shelters/index_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+    shelter_1 = Shelter.create(name: "Boulder shelter", city: "Aurora, CO", foster_program: false, rank: 9)
+    shelter_2 = Shelter.create(name: "Aurora shelter", city: "Aurora, CO", foster_program: false, rank: 9)
+    shelter_3 = Shelter.create(name: "Golder shelter", city: "Aurora, CO", foster_program: false, rank: 9)
+
+    visit "/admin/shelters"
+
+    expect(page).to have_content(shelter_1.name)
+    expect(page).to have_content(shelter_2.name)
+    expect(page).to have_content(shelter_3.name)
+    expect(shelter_3.name).to appear_before(shelter_2.name)
```
dcoleman21, line 14: This can also be wrapped in a `within` block

spec/features/pets_applications/show_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+        expect(page).to have_content("Lucille Bald")
+        click_button("Adopt this pet?")
+      end
+
+      within(".pets_submit") do
+        expect(page).to have_current_path("/pets_applications/#{application.id}")
+        expect(page).to have_content("Lucille Bald")
+      end
+      
+      within(".search_adopt") do
```
dcoleman21, line 53: Nice job using `within` blocks here

spec/models/pet_spec.rb
```diff
@@ -30,6 +32,23 @@
...
+        state: "Texas", zip_code: "75248", description: "I love animals!")
+        jane = Applicant.create!(name: "jane", street_address: "11234 Jane Street", city: "Dallas", 
+        state: "Texas", zip_code: "75248", description: "I love animals!")
+        shelter = Shelter.create!(name: "Aurora shelter", city: "Aurora, CO", foster_program: false, rank: 9)
+        pet_1 = shelter.pets.create!(adoptable: true, age: 1, breed: "sphynx", name: "Lucille Bald")
+        pet_2 = shelter.pets.create!(adoptable: true, age: 1, breed: "sphynx", name: "Barry")
+        application = PetsApplication.create!(applicant: dude, pet: pet_1, status: "Pending")
+        application_1 = PetsApplication.create!(applicant: dude, pet: pet_2)
+
+        expect(Pet.pets_with_app_status_by_sql(application)).to eq([pet_1, pet_2])
```
dcoleman21, line 48: Nice job testing your model method

PR #95, opened 2023-09-10
Ethan B + Tyler - Adopt Don't Shop

juliet-e:
Creating this PR to leave feedback for you!


app/controllers/admin/applications_controller.rb
```diff
@@ -0,0 +1,6 @@
+class Admin::ApplicationsController < ApplicationController
+  def show
+    @application = Application.find(params[:id])
+    @pets = @application.get_pets
```
juliet-e, line 4: It's okay to create @pets here but also keep in mind that you can call @application.get_pets in the view, too, and that's not a violation of MVC

juliet-e: We want to encapsulate any AR queries beyond CRUD in model methods. Then in the view, we stay away from doing any AR queries and instead just call those model methods that we made

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+      adoption_description: params[:adoption_description],
+      status: "Pending"
+    )
+    application.save
+    application.pets = pets
+
+    redirect_to "/applications/#{application.id}"
+  end
+
+  def add_pet
```
juliet-e, line 36: A more conventional/RESTful option here would be to put this functionality in a `create` action in the PetApplicationsController

Also, see note in application model but you can add a pet in a controller rather than encapsulating it at the model level. That's appropriate CRUD to leave in the controller!
In general, these AR methods are totally fine in a controller:
`.create`
`.save`
`.new`
`.update`
`.destroy`
`.find`
`.find_by`

It's also totally fine to call a model's associations in the controller, i.e.
`@application.pets`

app/models/application.rb
```diff
@@ -0,0 +1,35 @@
+class Application < ApplicationRecord
+  has_many :pet_applications, dependent: :destroy
+  has_many :pets, through: :pet_applications
+  validates :name, presence: true
+  validates :address, presence: true
+  validates :description, presence: true
+  validates_inclusion_of :status, in: ["In Progress", "Pending", "Accepted", "Rejected"]
+
+  def get_pets
```
juliet-e, line 9: No need to encapsulate this association in a model method. When we say `has_many :pets`, what Rails is doing under the hood is actually creating an instance method  called `pets` that will run a query to find all pets associated with this application. So we can call this `.pets` as if it's just another model method but it represents, specifically, our associations. 

You can just call `application.pets` in a controller or view

app/models/application.rb
```diff
@@ -0,0 +1,35 @@
...
+    pets.pluck(:name)
+  end
+
+  def pet_link(pet_name)
+    pet = Pet.find_by(name: pet_name)
+    "/pets/#{pet.id}" if pet
+  end
+
+  def add_pet(pet_id)
+    pet = Pet.find(pet_id)
```
juliet-e, line 23: this is purely CRUD and could live in the controller rather than being in a model method

app/models/application.rb
```diff
@@ -0,0 +1,35 @@
...
+
+    self.pets << pet
+  end
+
+  def has_pets?
+    pets.exists?
```
juliet-e, line 29: These other model methods without comments (and this one `has_pets?`) are great options to keep in the model!

app/models/pet.rb
```diff
@@ -10,4 +12,8 @@ def shelter_name
   def self.adoptable
     where(adoptable: true)
   end
+
+  def self.search(name)
+    where("name ILIKE ?", "%#{name}%")
```
juliet-e, line 17: Note that this method also exists in the ApplicationRecord file. Since every model inherits from ApplicationRecord, you have access to this method in any model even without declaring it again! You can try this out just by deleting this method and noticing that your application still works, because now your code is going to that parent method declared in ApplicationRecord instead!

app/models/shelter.rb
```diff
@@ -16,6 +16,17 @@ def self.order_by_number_of_pets
       .order("pets_count DESC")
   end
 
+  def self.order_by_reverse_alphabetical
+    find_by_sql("SELECT * FROM shelters ORDER BY name DESC")
+  end
+
+  def self.shelters_with_pending_applications
+    joins(pets: { pet_applications: :application })
```
juliet-e, line 24: This is a fun one for feedback! First I'm going to talk through what you're doing here and then I'll show you an easier option (but this one works great, too!)
First, you're joining the shelters table to pets, and then joining the pet_applications table to that temporary table, and then finally joining the applications table on top of that. You're doing a multi-part join here. This temporary table is HUGE and has all the data from shelters, pets, pet_applications and applications - basically every table we care about in this project. 

From there, we're filtering the data down to those rows where the application `status` column is `Pending`. We had to do this complex joining in the first step, because the shelter table alone didn't have any information about applications, so we had to do all those joins in order to arrive at applications, where this data lives. 

Because shelters have many pets, and pets can belong to many applications, this temporary joins table probably has MANY duplicate shelters. Anytime a pet from this shelter has an application associated to it, this shelter appears again. We want all the names of the shelters, but we can remove the duplicates. 

Then we're just plucking the name column to end up with an array of shelter names.

Now, a very cool AR trick here is to use associations in order to avoid having to do multi-joins like this. I call it association "mooching". We know that the `Pet` model has these asssociations: `has_many pet_applications` and `has_many :applications, through: :pets`. Since `Pet` belongs to `Shelter`, `Shelter` has permission to mooch off of these associations that its child object has. You can, therefore, add these additional associations to your Shelter model in order to leverage the associations Shelter will get from Pet:
`has_many :applications, through: :pets`

We could include `pet_applications` too but what we care about with this story is `application` status so we can just cut to the chase and go right to applications. What this says to Rails is: "Since Pet is my child object, and Pet has many applications, then I, the shelter, have many applications, too"

So now, you can do this for this query:
`self.applications.where(applications: { status: 'Pending' }).distinct.pluck(:name)`
You can also get rid of the `self` and start just with `applications`

app/views/admin/applications/show.html.erb
```diff
@@ -0,0 +1,37 @@
...
+        <td><%= pet.breed %></td>
+      </tr>
+      <tr>
+        <td>Shelter ID: </td>
+        <td><%= pet.shelter_id %></td>
+      </tr>
+    </table>
+    <% if @application.pet_status(pet) == nil %>
+      <br>
+      <%= button_to "Approve Pet", "/admin/applications/#{@application.id}?pet_id=#{pet.id}&status=approved", method: :patch, data: {turbo: false} %>
```
juliet-e, line 30: Luckily, Turbo doesn't mess with buttons so you can leave the Turbo parameter out and it shouldn't break anything! Turbo needs to be disabled only for links and forms

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,70 @@
+<%= render partial: '/layouts/application_header', locals: { application: @application } %>
```
juliet-e, line 1: Nice work including a partial!

config/routes.rb
```diff
@@ -36,4 +36,15 @@
   get "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinary_offices#veterinarians"
   get "/veterinary_offices/:veterinary_office_id/veterinarians/new", to: "veterinarians#new"
   post "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinarians#create"
+
+  post "/applications", to: "applications#create"
+  get "/applications/new", to: "applications#new"
+  get "/applications/pets/:id", to: "pets#show"
+  get "/applications/:id", to: "applications#show"
+  patch "/applications/:id", to: "applications#update"
+  post "/applications/:id/pets", to: "applications#add_pet"
```
juliet-e, line 45: Nice work with the routes! This is the only route that could be made more RESTful (as mentioned previously) by using "pet_applications#create"

spec/features/admin/applications/show_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  describe "When I visit /admin/applications/:id" do
+    it "For every pet on the application, I see an approval button for that pet" do
+      visit "/admin/applications/#{@application_1.id}"
+
+      expect(page).to have_content(@application_1.name)
+      expect(page).to have_content(@application_1.address)
+      expect(page).to have_content(@application_1.description)
+
+      within("#pet-#{@pet_1.id}") do
```
juliet-e, line 21: Great work using within blocks!!

spec/features/admin/applications/show_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+      visit "/admin/applications/#{@application_2.id}"
+      
+      within("#pet-#{@pet_1.id}") do
+        expect(page).to have_button("Approve Pet")
+        expect(page).to have_button("Reject Pet")
+        expect(page).to_not have_content("Pet status:")
+      end
+    end
+  end
+end
```
juliet-e, line 86: Excellent, robust testing!

spec/features/admin/shelters/index_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+    expect(page).to have_content(@shelter_2.name)
+    expect(page).to have_content(@shelter_3.name)
+  end
+
+  it "lists all the shelter names in reverse alphabetical order" do
+    visit "/admin/shelters"
+
+    end_of = find("#shelter-#{@shelter_2.id}")
+    mid_of = find("#shelter-#{@shelter_3.id}")
+    start_of = find("#shelter-#{@shelter_1.id}")
```
juliet-e, line 26: This works, and is an okay way to test order, but something that might be even easier is skipping these `find` methods and just testing:
`expect(@shelter_2).to appear_before(@shelter_3)`
since you still have access to those instance variables from the before :each!

spec/features/applications/new_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    new_app = Application.all.first
+
+    expect(page).to have_current_path("/applications/#{new_app.id}")
+    expect(page).to have_content("Tyler Blackmon")
+    expect(page).to have_content("1234 Street Address, Colorado Springs, CO, 80922")
+    expect(page).to have_content("I have a really good description so you should let me have a pet!")
+    expect(page).to have_content("In Progress")
+  end
+
+  it "sends an error message if any form fields are not filled in" do
```
juliet-e, line 36: Nice sad path testing!

spec/models/application_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+      it "returns the current adoption status of a pet on an individual application" do
+        PetApplication.create!(pet_id: @pet.id, application_id: @application_2.id)
+        PetApplication.create!(pet_id: @pet_2.id, application_id: @application_1.id, status: "Approved")
+
+        expect(@application_1.pet_status(@pet_2)).to eq("Approved")
+        expect(@application_2.pet_status(@pet)).to be nil
+      end
+    end
+  end
+end
```
juliet-e, line 73: Great work testing this functionality at both the unit (model) level and integration (feature) level!

PR #97, opened 2023-09-10
Will + Blake: Adopt Don't Shop

juliet-e:
Creating this PR to leave you feedback!


app/controllers/admin/applications_controller.rb
```diff
@@ -0,0 +1,22 @@
+class Admin::ApplicationsController < ApplicationController
+  def show
+    @application = Application.find(params[:id])
+  end
+
+  def update
+    @pet_application = PetApplication.find(params[:id])
+    @application = @pet_application.application
+    @pet = @pet_application.pet
```
juliet-e, line 9: It's okay to make these instance vars here, but keep in mind that you can call `@pet_application.application` and `@pet_application.pet` from the view too! You're not violating MVC by doing that

app/controllers/admin/shelters_controller.rb
```diff
@@ -0,0 +1,6 @@
+class Admin::SheltersController < ApplicationController
+  def index
+    @shelters = Shelter.order_desc_alphabetical
```
juliet-e, line 3: Nice work encapsulating this logic at the model level!!

app/controllers/pet_applications_controller.rb
```diff
@@ -0,0 +1,13 @@
+class PetApplicationsController < ApplicationController
+  def create
```
juliet-e, line 2: Nice work pulling this functionality into a 3rd Joins controller! This is the most RESTful option.

app/models/pet.rb
```diff
@@ -10,4 +13,12 @@ def shelter_name
   def self.adoptable
     where(adoptable: true)
   end
+
+  def self.adoptable_search(search_params)
+    adoptable.search(search_params)
+  end
+
+  def update_adoptable_status
+    update(adoptable: false) if pet_applications.any? { |pa| pa.approval == true }
```
juliet-e, line 22: These are okay model methods to put some Ruby logic in (as you've done) because there isn't an easier way to do this in AR

app/models/shelter.rb
```diff
@@ -31,4 +31,12 @@ def alphabetical_pets
   def shelter_pets_filtered_by_age(age_filter)
     adoptable_pets.where("age >= ?", age_filter)
   end
+
+  def self.order_desc_alphabetical
+    find_by_sql("SELECT * FROM shelters ORDER BY name DESC")
+  end
+
+  def self.pending_applications
+    joins(pets: :applications).where("applications.status = 'Pending'").distinct
```
juliet-e, line 40: Great work! This is a nice, concise joins method. 

Another nifty option: Allow Shelter to mooch off the associations that Pet has set up. The Pet model has the associations: `has_many :applications, through: pet_applications`. You can build off of associations in Rails, so in your Shelter model, because Shelter `has_many :pets`, you can also add `has_many :applications, through: :pets`. Then, you can access `applications` as an attribute of a shelter! It will do this multi-part joining to the applications table under the hood!

app/views/admin/applications/show.html.erb
```diff
@@ -0,0 +1,50 @@
...
+      <th>Reject</th>
+      <th>Approval Status</th>
+    </tr>
+  </thead>
+  <tbody>
+    <% @application.pet_applications.each do |pet_application| %>
+      <tr>
+        <td><%= pet_application.pet.name %></td>
+        <td>
+          <% if pet_application.approval == nil && pet_application.pet.adoptable != false %>
```
juliet-e, line 24: Having a somewhat complex conditional decide whether you render a form is often the case. This is okay IMO and not `too complex`

app/views/admin/applications/show.html.erb
```diff
@@ -0,0 +1,50 @@
...
+        </td>
+        <td>
+          <% if pet_application.approval == nil %>
+            <%= form_tag admin_application_path(@application, id: pet_application.id), method: :patch do %>
+            <%= submit_tag 'Reject' %>
+            <% end %>
+          <% end %>
+        </td>
+        <td>
+          <% if pet_application.approval == true || pet_application.pet.adoptable == false %>
```
juliet-e, line 38: When you have a complex condition and it decides whether to render just simple _text_ however, I think you could move this conditional into a model method. It is presentation logic, but for something simple like this, I think it's not an MVC violation to put this in the model so that a PetApplication method can be called to determine what text is displayed here. Then, you could just render `pet_application.status_text` to get the label that's necessary

app/views/admin/applications/show.html.erb
```diff
@@ -0,0 +1,50 @@
...
+            <%= "APPROVED" %>
+          <% elsif pet_application.approval == false %>
+            <%= "REJECTED" %>
```
juliet-e, line 41: These labels are just text and don't need to be wrapped in ERB claws <%= %>

app/views/admin/shelters/index.html.erb
```diff
@@ -0,0 +1,20 @@
+<%= render partial: 'admin/header', locals: { name: "Shelters" } %>
```
juliet-e, line 1: Great work using a partial!

app/views/admin/shelters/index.html.erb
```diff
@@ -0,0 +1,20 @@
...
+<h1>All Shelters</h1>
+<% @shelters.each do |shelter| %>
+  <div id="shelter-<%= shelter.id %>">
+    <h3><%= shelter.name %></h3>
+    <p>Created at: <%= shelter.created_at %></p>
+    <%= link_to "Update #{shelter.name}", "/shelters/#{shelter.id}/edit" %>
+    <%= link_to "Delete #{shelter.name}", "/shelters/#{shelter.id}", method: :delete %>
```
juliet-e, line 8: Just a heads up that this `delete` doesn't work on the deployed site because you need to disable Turbo! Or could swap this for a button if you don't want to deal with Turbo nonsense

app/views/applications/new.html.erb
```diff
@@ -0,0 +1,22 @@
+<h1>New Application</h1>
+<%= form_with url: "/applications", method: :post, local: true do |f| %>
```
juliet-e, line 2: Make sure to use data: {turbo: false} instead of local:true in the future, just in case, to avoid Turbo issues.

app/views/shelters/index.html.erb
```diff
@@ -16,3 +16,4 @@
     <%= link_to "Delete #{shelter.name}", "/shelters/#{shelter.id}", method: :delete %>
```
juliet-e, line 16: All these link_to methods without disabling Turbo might cause a problem

config/routes.rb
```diff
@@ -36,4 +42,8 @@
   get "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinary_offices#veterinarians"
   get "/veterinary_offices/:veterinary_office_id/veterinarians/new", to: "veterinarians#new"
   post "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinarians#create"
+
+  get "/admin/shelters", to: "admin/shelters#index"
+  get "/admin/applications/:id", to: "admin/applications#show"
+  patch "/admin/applications/:id", to: "admin/applications#update", as: "admin_application"
```
juliet-e, line 48: Excellent RESTful routing!

spec/features/admin/applications/show_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+
+      expect(page).to have_button('Approve', count: 2)
+
+      within("tr:contains('#{@pet1.name}')") do
+        click_button 'Approve'
+      end
+
+      expect(page).to have_current_path("/admin/applications/#{@application2.id}")
+
+      within("tr:contains('#{@pet1.name}')") do
```
juliet-e, line 31: Nice use of within blocks!

spec/models/shelter_spec.rb
```diff
@@ -41,6 +41,25 @@
...
+        application = Application.create!(applicant_name: "Thomas Jefferson", street_address: "123 Main St.", city: "Boston", state: "MA", zip_code: "12345", description: "I'm on a fiver", status: "In Progress")
+        application.pets << @pet_1
+
+        expect(Shelter.pending_applications).to_not eq([@shelter_1])
+
+        application.update(status: "Pending")
+
+        expect(Shelter.pending_applications).to eq([@shelter_1])
+      end
+    end
```
juliet-e, line 62: Nice work testing your model methods at the unit (model) and integration (feature) level!

PR #98, opened 2023-09-10
Reed + Dylan Adopt Don't Shop

juliet-e:
Creating this PR to leave you feedback!


app/controllers/admin/pet_applications_controller.rb
```diff
@@ -0,0 +1,16 @@
+class Admin::PetApplicationsController < ApplicationController
+    def update
+        pet_application = PetApplication.where(["application_id = ? and pet_id = ?", "#{params[:id]}", "#{params[:pet_id]}"])
```
juliet-e, line 3: Since we can be fairly sure there is only one Pet App with the given pet_id and app_id, I think `find_by` would work here and then you wouldn't have to unwrap the record out of an array!

juliet-e: Although it looks like the `update` action was smart enough to do the unwrapping for you. But it's good to keep in mind that `where` will return an ARRAY rather than a single record

app/controllers/admin/shelters_controller.rb
```diff
@@ -0,0 +1,6 @@
+class Admin::SheltersController < ApplicationController
+  def index
+    @shelters = Shelter.order_by_reverse_alphabetical
+    @pending_apps = Shelter.pending_apps.alphabetical_order
```
juliet-e, line 4: Nice work encapsulating this logic at the model level rather than doing it all here!

app/controllers/admin/pet_applications_controller.rb
```diff
@@ -0,0 +1,16 @@
+class Admin::PetApplicationsController < ApplicationController
```
juliet-e, line 1: Nice work using the Admin namespacing and pulling this out into a Joins controller! This is the most RESTful option!

app/controllers/pet_applications_controller.rb
```diff
@@ -0,0 +1,10 @@
+class PetApplicationsController < ApplicationController
+    def create
```
juliet-e, line 2: Great work using a joins controller here as well to handle creating the joins record

app/models/shelter.rb
```diff
@@ -4,6 +4,8 @@ class Shelter < ApplicationRecord
   validates :city, presence: true
 
   has_many :pets, dependent: :destroy
+  has_many :pet_applications, through: :pets
+  has_many :applications, through: :pet_applications
```
juliet-e, line 8: Excellent job leveraging the associations that pet has set up in the shelter model!! This makes the join to applications muuuuch easier

app/models/shelter.rb
```diff
@@ -16,6 +18,14 @@ def self.order_by_number_of_pets
       .order("pets_count DESC")
   end
 
+  def self.order_by_reverse_alphabetical
+    find_by_sql("SELECT * FROM shelters ORDER BY name DESC")
+  end
+
+  def self.pending_apps
+    joins(:applications).where("status = 'Pending'")
```
juliet-e, line 26: With the associations you have set up, `applications` is now an attribute on shelters so you can skip the joins method here and just start with `applications.where`

app/views/admin/applications/show.html.erb
```diff
@@ -0,0 +1,20 @@
...
+  <div id="pet-<%= pet.id %>">
+    <%= link_to "#{pet.name}", "/pets/#{pet.id}" %><br>
+    <% if @application.find_if_adoptable(pet.id) == false && @application.find_pet_application_status(pet.id) == "Pending" %>
+      <p>This pet has been approved for adoption</p><br>
+      <%= button_to "Reject Application", "/admin/applications/#{@application.id}?pet_id=#{pet.id}&button=reject", method: :patch %>
+    <% elsif @application.find_pet_application_status(pet.id) == "Approved" %>
+      <p>Application Approved</p>
+    <% elsif @application.find_pet_application_status(pet.id) == "Rejected" %>
+      <p>Application Rejected</p>
+    <% elsif @application.find_pet_application_status(pet.id) == "Pending" %>
```
juliet-e, line 14: This view is going to be pretty complex, so none of these refactors make a huge world of a different but here are some ideas:

- First IF branch: `@application.find_pet_application_status(pet.id) == "Pending"`
- - - inside this if branch, check for `@application.find_if_adoptable(pet.id) == false` to indicate this pet isn't available
- - - otherwise add buttons
- ELSE
- - - render the appropriate text based on Approved/Rejected/Pending. You could even have this logic in a model method and call this method here because all you're rendering is text. Something like
- - - `<%= @application.find_pet_app_status_text_label(pet.id) %>`  which will return Approved or Rejected

app/views/admin/shelters/index.html.erb
```diff
@@ -0,0 +1,18 @@
+<%= render partial: "layouts/shelter_render" %>
+
+<% @shelters.each do |shelter| %>
+  <div id="shelter-<%= shelter.id %>">
+    <h3><%= shelter.name %></h3>
+    <p>Created at: <%= shelter.created_at %></p>
+    <%= link_to "Update #{shelter.name}", "/shelters/#{shelter.id}/edit" %>
+    <%= link_to "Delete #{shelter.name}", "/shelters/#{shelter.id}", method: :delete %>
```
juliet-e, line 8: This won't work without disabling Turbo! Remember that if you use a link_to for anything other than a GET request, Turbo will interfere. Be sure to include `{ turbo_method: :delete }` or just use a button instead

app/views/shelters/index.html.erb
```diff
@@ -1,4 +1,4 @@
-<h1>All Shelters</h1>
+<%= render partial: "layouts/shelter_render" %>
```
juliet-e, line 1: Nice work using Partials!

config/routes.rb
```diff
@@ -36,4 +36,15 @@
...
+  get "/applications", to: "applications#index"
+  get "/applications/new", to: "applications#new"
+  post "/applications", to: "applications#create"
+  get "/applications/:id", to: "applications#show"
+  patch "/applications/:id", to: "applications#update"
+  post "applications/:id/pet_applications", to: "pet_applications#create"
+
+  get "/admin/shelters", to: "admin/shelters#index"
+  get "/admin/applications/:id", to: "admin/applications#show"
+  patch "/admin/applications/:id", to: "admin/pet_applications#update"
```
juliet-e, line 49: Excellent RESTful routing!

spec/features/admin/applications/show_spec.rb
```diff
@@ -0,0 +1,229 @@
...
+  it "can reject pets for adoption" do
+    visit "/admin/applications/#{@applicant_1.id}"
+
+    expect(page).to have_button("Reject Application")
+
+    click_button("Reject Application")
+
+    expect(current_path).to eq("/admin/applications/#{@applicant_1.id}")
+    
+    within("#pet-#{@pet_1.id}") do
```
juliet-e, line 55: Nice use of within blocks!

spec/features/applications/create_spec.rb
```diff
@@ -0,0 +1,123 @@
...
+        visit "/applications/new"
+
+        fill_in(:name, with: "Billy")
+        fill_in(:city, with: "Rainbow Road")
+        click_button("Submit")
+
+        expect(current_path).to eq("/applications/new")
+        expect(page).to have_content("Error: Street address can't be blank, State can't be blank, Zip code can't be blank")
+    end
+end
```
juliet-e, line 123: Very robust testing of all the sad paths!

spec/models/shelter_spec.rb
```diff
@@ -67,5 +67,67 @@
...
+      end
+    end
+
+    it "orders shelters by name" do
+
+      ordered_shelters = Shelter.alphabetical_order.to_a
+
+      expected_order = [@shelter_1, @shelter_3, @shelter_2]
+
+      expect(ordered_shelters).to eq(expected_order)
```
juliet-e, line 108: While it seems simple, this is a nice and robust test for this functionality! For model specs, you just want to make sure that at the lowest data level, you're testing the happy path (this test) and any additional edge cases that might be relevant. Sometimes, the logic is complex enough that additional tests are appropriate. This is a simple enough query that we don't need to handle alternative or sad paths (although you could if you tested that nothing breaks if you call this order method when there are no shelters... but that's overkill IMO)

The pending apps method has a bit more complexity to it and so you had to create more test data. You're testing the happy path, but I think you could make the test more robust by verifying that the method only returns ONE shelter back: `expect(shelters_with_pending_apps.length).to eq(1)` --> you could add this right after your assertion to make sure that not only does this return include that one shelter, but it's the only shelter that's given back. Your method is not erroneously returning non-pending apps.

PR #99, opened 2023-09-10
Eliza and Edward - Adopt Don't Shop

juliet-e:
Creating this PR to leave you some feedback!


app/controllers/admin_controller.rb
```diff
@@ -0,0 +1,21 @@
+class AdminController < ApplicationController
```
juliet-e, line 1: Another option here for dealing with an Admin controller:

Often, our admin controllers deal with resources in the same way our regular controllers do - they handle shelters, and applications, and other records in our db. So, in order to make an application more extendable, it's usually easier to not having one single AdminController, but to have an admin *directory* and put new controllers inside of it depending on what functionality the admin needs to do. For example, for the admin shelter functionality, you can create a SheltersController inside the `admin` directory and it won't collide with the SheltersController that already exists outside the `admin` directory. You'll see more on this in your Advanced Routing class!

juliet-e: This also allows better SRP adherence. Note how in this controller, you're dealing with lots of different things -- shelters, applications, etc. It's best for controllers to handle just one specific resource

app/controllers/admin_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+    @shelters = Shelter.alphabetical_sort
+    @shelters_pending = Shelter.pending_applications
+  end
+
+  def show
+    @application = Application.find(params[:id])
+  end
+
+  def update
+    relationship = ApplicationPet.where("pet_id = ? AND application_id = ?", params[:pet_id], params[:application_id])
```
juliet-e, line 12: Since we can be reasonably sure there should be only one ApplicationPet with this given id combo, I'd generally stick with `find_by` instead of `where` because `where` returns an array of records rather than a single record

app/controllers/application_pet_controller.rb
```diff
@@ -0,0 +1,9 @@
+class ApplicationPetController < ApplicationController
+  def create
```
juliet-e, line 2: Great work pulling this action into its own controller for joining Applications with Pets!

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,44 @@
+class ApplicationsController < ApplicationController
+  def show
+    @application = Application.find(params[:id])
+    if params[:search] == "pet" && params[:pet_name].present?
+      search_query = params[:pet_name]
+      @searched_pets = Pet.where("name ILIKE ?", "%#{search_query}%")
```
juliet-e, line 6: This is logic we'd prefer to keep encapsulated at the model level. Also, take a look at the ApplicationRecord file -- this model method is already created for you! Since every model method inherits from ApplicationRecord, the `.search` method is available for you in every model! The beauty of a brownfield project is finding things you can use that other devs wrote

app/models/pet.rb
```diff
@@ -10,4 +12,8 @@ def shelter_name
   def self.adoptable
     where(adoptable: true)
   end
+
+  def find_app_relationship(application_id)
+    ApplicationPet.where("pet_id = ? AND application_id = ?", id, application_id).first
```
juliet-e, line 17: This is another case where you could do `find_by`

app/models/shelter.rb
```diff
@@ -31,4 +31,12 @@ def alphabetical_pets
   def shelter_pets_filtered_by_age(age_filter)
     adoptable_pets.where("age >= ?", age_filter)
   end
+
+  def self.alphabetical_sort
+    Shelter.find_by_sql("SELECT * FROM shelters ORDER BY name DESC")
+  end
+
+  def self.pending_applications
+    Shelter.joins(pets: :applications).where("applications.status = ?", "Pending")
```
juliet-e, line 40: This join looks great!

Another nifty option: Allow Shelter to mooch off the associations that Pet has set up. The Pet model has the associations: `has_many :applications, through: pet_applications`. You can build off of associations in Rails, so in your Shelter model, because Shelter `has_many :pets`, you can also add `has_many :applications, through: :pets`. Then, you can access `applications` as an attribute of a shelter! It will do this multi-part joining to the applications table under the hood!

app/views/admin/_button.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= form_with url: url, method: method do |f| %>
+    <%= f.submit button_text %>
+<% end %>
```
juliet-e, line 3: Nice use of partials! I wonder, though, whether the number of local variables you have to pass in for such a small partial makes this maybe not worth the effort. Partials are great, but you can really see their power when you use them to replace a more complex form with multiple fields

app/views/admin/show.html.erb
```diff
@@ -0,0 +1,35 @@
...
+
+<p>Zip Code: <%= @application.zip_code %></p>
+
+<p>Description: <%= @application.owner_description %></p>
+
+<p id="pets">Pets: 
+  <% @application.pets.each do |pet| %>
+      <br><%= link_to "#{pet.name}", "/pets/#{pet.id}" %>
+      <% if pet.find_app_relationship(@application.id).status.nil? %>
+        <%= render partial: 'button', locals: {
```
juliet-e, line 17: Here's what I mean by the partial not being worth the effort for this spot: that's a lot of local vars to pass in for one little button. Is there another spot in your application where a partial could've been more powerful?

app/views/shelters/_form.html.erb
```diff
@@ -0,0 +1,15 @@
+<%= form_with url: path, method: method do |f| %>
```
juliet-e, line 1: THIS is a much more powerful partial!!!

app/views/welcome/index.html.erb
```diff
@@ -0,0 +1,5 @@
+<h2>Hello World</h2>
+<p>Adopt, don't shop!</p>
+<p>
+  The time is now: <%= Time.now %>
```
juliet-e, line 4: From now on until eternity, I will come to your landing page to find the current time. Lovely.

juliet-e: who needs clocks?!

config/routes.rb
```diff
@@ -36,4 +37,15 @@
   get "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinary_offices#veterinarians"
   get "/veterinary_offices/:veterinary_office_id/veterinarians/new", to: "veterinarians#new"
   post "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinarians#create"
+
+  get "/applications/new", to: "applications#new"
+  get "/applications/:id", to: "applications#show"
+  post "/applications", to: "applications#create"
+  patch "/applications/:id", to: "applications#update"
+
+  post "/application_pet", to: "application_pet#create"
```
juliet-e, line 46: This is extremely minor but in a joins controller, it's conventional to pluralize the second record, i.e. `application_pets` (while application stays singular)

config/routes.rb
```diff
@@ -36,4 +37,15 @@
   get "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinary_offices#veterinarians"
   get "/veterinary_offices/:veterinary_office_id/veterinarians/new", to: "veterinarians#new"
   post "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinarians#create"
+
+  get "/applications/new", to: "applications#new"
```
juliet-e, line 41: Great RESTful routing here!

config/routes.rb
```diff
@@ -36,4 +37,15 @@
...
+  get "/applications/:id", to: "applications#show"
+  post "/applications", to: "applications#create"
+  patch "/applications/:id", to: "applications#update"
+
+  post "/application_pet", to: "application_pet#create"
+
+  get "/admin/shelters", to: "admin#index"
+  get "/admin/applications/:id", to: "admin#show"
+  patch "/admin/applications/:id", to: "admin#update"
```
juliet-e, line 50: See my comment in the AdminController but here is where you could make use of using an admin directory. i.e.
```
  get "/admin/shelters", to: "admin/shelters#index"
  get "/admin/applications/:id", to: "admin/applications#show"
  patch "/admin/applications/:id", to: "admin/applications#update"
```

spec/features/admin/index_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+        it "and In this section I see the name of every shelter that has a pending application" do
+          @application.pets << @pet_1
+
+          visit "/admin/shelters"
+
+          expect(page).to have_content("Shelters with Pending Applications")
+
+          within("#pending-applications") do
+            expect(page).to have_content("The Shelter")
+            expect(page).not_to have_content("A Shelter")
```
juliet-e, line 66: What is "A Shelter"? Did you mean "City Shelter" or "Zoo Shelter"? Testing for those would make this test more robust because right now I don't have confirmation that the other created shelters don't show up

spec/features/admin/show_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+
+        visit "/admin/applications/#{@application_1.id}"
+
+        @application_1.pets.each do |pet|
+          expect(page).to have_button("Reject #{pet.name}")
+          click_button("Reject #{pet.name}")
+
+          expect(page).to have_no_button("Reject #{pet.name}")
+          expect(page).to have_content("#{pet.name} has been rejected")
+        end
```
juliet-e, line 82: Nice robust testing here!

spec/features/applications/show_spec.rb
```diff
@@ -0,0 +1,228 @@
...
+            click_button("Submit Application")
+
+            expect(current_path).to eq("/applications/#{@application.id}")
+
+            within("#status") do
+              expect(page).to have_content("Pending")
+            end
+
+            within("#pets") do
+              expect(page).to have_content("Cheesecake")
```
juliet-e, line 128: Nice use of within blocks!

PR #148, opened 2023-10-26
Anthea/Miranda

juliet-e:
Creating this PR to leave you feedback on your project! :)


app/controllers/admin/application_pets_controller.rb
```diff
@@ -0,0 +1,19 @@
+class Admin::ApplicationPetsController < ApplicationController
+
+  def update
```
juliet-e, line 3: It's okay to keep this controller action in your other non-admin ApplicationPets controller if you want! The route from the button to approve a pet doesn't need to be prefixed with `/admin` so it could go in the regular ApplicationPets controller. But this is not "wrong" ! Just another option

app/controllers/admin/application_pets_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+    application = Application.find(params[:application_id])
+    pet = Pet.find(params[:pet_id])
+    match = ApplicationPet.find_by(pet_id: pet.id, application_id: application.id)
+    if params[:approve] == "true"
+      match.approve
+      pet.approve
```
juliet-e, line 9: Does a pet need approving in addition to the ApplicationPet joins record status? What does pet approval do?

juliet-e: I wonder if you could reduce the complexity here by having a general `update_status` method in the PetApplication model and you could pass the `params[:approve]` in as a parameter. That way you could push the complexity down into a model method and keep the controller a bit slimmer. This isn't an MVC violation but just a design idea!

app/controllers/admin/shelters_controller.rb
```diff
@@ -0,0 +1,6 @@
+class Admin::SheltersController < ApplicationController
+  def index 
+    @shelters = Shelter.order_by_reverse_alphabetical
+    @shelters_with_pending_applications = Shelter.shelters_with_pending_applications
```
juliet-e, line 4: Beautiful encapsulation of business logic into model methods!

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,41 @@
+class ApplicationsController < ApplicationController
+  def show
+    @application = Application.find(params[:id])
+    @pets = @application.pets
+    if params[:search] == "pet" && params[:pet_name].present? 
+      pet_name = params[:pet_name]
+      @search_pets = Pet.where("name ILIKE ?", "%#{pet_name}%")
+    else 
+      @search_pets = []
```
juliet-e, line 9: Instead of having an empty instance variable, it's slightly more conventional to have a check in the view, something like: `if @search_pets.present?`

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,41 @@
+class ApplicationsController < ApplicationController
+  def show
+    @application = Application.find(params[:id])
+    @pets = @application.pets
+    if params[:search] == "pet" && params[:pet_name].present? 
+      pet_name = params[:pet_name]
+      @search_pets = Pet.where("name ILIKE ?", "%#{pet_name}%")
```
juliet-e, line 7: We don't want a search query with .where in the controller, as this would be an MVC violation. Controllers should only use the following CRUD activerecord methods: .create, .update, .destroy, .find, .find_by, .all

Also, this is a brownfield project so this method is actually already written for us! Check out the search method in the ApplicationRecord file. Alll models inherit from this parent class, so you can actually do `Pet.search(pet_name)` here and it would direct to that method

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+      name: params[:name],
+      street_address: params[:street_address],
+      city: params[:city],
+      state: params[:state],
+      zip_code: params[:zip_code],
+      description: params[:description],
+      status: "In Progress"
+    })
+
+    if application.save
```
juliet-e, line 27: In order for `application.save` to work properly for error handling, line 17 should have Application.new (i.e. you just instantiate the object on line 17 with .new and then .save on line 27 tries to save it to the db)

app/controllers/pets_controller.rb
```diff
@@ -16,7 +16,7 @@ def new
   end
 
   def create
-    pet = Pet.new(pet_params)
+    pet = Pet.create(pet_params)
```
juliet-e, line 19: This should be kept as .new + .save! .create does both of the but when we want to capture any errors that can happen when saving, we break them up into `.new` and `.save` in order to capture the error case where a record isn't created properly and therefore can't be saved

app/controllers/application_pets_controller.rb
```diff
@@ -0,0 +1,10 @@
+class ApplicationPetsController < ApplicationController
```
juliet-e, line 1: Awesome RESTful controller organization here!

app/models/application.rb
```diff
@@ -0,0 +1,28 @@
...
+  validates :zip_code, presence: true
+  validates :description, presence: true
+  validates :status, presence: true
+
+  def address_join
+    "#{street_address}, #{city}, #{state}, #{zip_code}"
+  end
+
+  def change_status
+    check_values = application_pets.map{|pet| pet.status}
```
juliet-e, line 20: It's okay to use Ruby here because the ActiveRecord equivalent here is pretty hard and not necessarily more elegant!

I'm curious if the nil check here will cause applications to be prematurely considered "Rejected" if none of their application_pets have been approved. The status column is probably defaulting to nil, that's why I'm wondering! Is the nil check necessary? THere is probably a nil edge case I'm not considering

app/models/pet.rb
```diff
@@ -10,4 +12,16 @@ def shelter_name
   def self.adoptable
     where(adoptable: true)
   end
+
+  def self.search(name)
```
juliet-e, line 16: Nice work encapsulating this at the model level, but it doesn't seem like you used this in your controller. Also, check out ApplicationRecord -- it's already written for you!

app/controllers/admin/application_pets_controller.rb
juliet-e: I see now that pet.approve is changing the adoptable flag! My bad

app/models/pet.rb
```diff
@@ -10,4 +12,16 @@ def shelter_name
...
+    where("name ILIKE ?", "%#{name}%")
+  end
+
+  def approve
+    self.update(adoptable: false)
+  end
+
+  def application_approved?
+    applications.map{|app| app.status}.include?("Approved")
```
juliet-e, line 25: Similar to a previous comment, the more Railsy/AR way to do this would be `applications.where(status: "Approved").empty?` or you would actually probably want .present? so that if it's false, that means none are approved

app/models/shelter.rb
```diff
@@ -31,4 +32,13 @@ def alphabetical_pets
   def shelter_pets_filtered_by_age(age_filter)
     adoptable_pets.where("age >= ?", age_filter)
   end
+
+  # RAW SQL
+  def self.order_by_reverse_alphabetical
```
juliet-e, line 37: These look good, just make sure there are model tests for them as well!! I believe model tests for these methods are missing

app/models/shelter.rb
```diff
@@ -31,4 +32,13 @@ def alphabetical_pets
...
+    find_by_sql("SELECT * FROM shelters ORDER BY name DESC")
+  end
+
+  def self.shelters_with_pending_applications
+    Shelter.joins(pets: { application_pets: :application }).where(applications: { status: 'Pending' }).distinct.pluck(:name)
```
juliet-e, line 42: Because you have the additional association to applications, you can also do `Shelter.joins(:applications)` directly and you should see the SQL that is used under the hood is doing the join through pets automatically  because you've created this shortcut with model associations

app/views/admin/applications/show.html.erb
```diff
@@ -0,0 +1,17 @@
...
+
+<h3>All Pets Applied For:</h3>
+<% @application.application_pets.each do |app_p| %>
+    <h3><%= app_p.pet.name %></h3>
+    <h4><%= "** Approved!**" if app_p.approved(@application.id) %></h4>
+    <h4><%= "Pet has been approved on a different application!" if app_p.pet.application_approved? && !app_p.approved(@application.id) %></h4>
+    <h4><%= "** Rejected **" if app_p.rejected(@application.id) %></h4>
+    <p> <%= "Status: Adoptable" if app_p.pet.adoptable%> <%= "Status: Not Adoptable" if !app_p.pet.adoptable && !app_p.approved(@application.id)%></p> 
+    <%= button_to "Approve #{app_p.pet.name}", {controller: "application_pets", action: "update", application_id: @application.id, pet_id: app_p.pet.id, approve: true}, method: :patch if !app_p.pet.application_approved? && !app_p.approved(@application.id) && !app_p.rejected(@application.id)%>
+    <%= button_to "Reject #{app_p.pet.name}", {controller: "application_pets", action: "update", application_id: @application.id, pet_id: app_p.pet.id, approve: false}, method: :patch if !app_p.pet.application_approved? && !app_p.approved(@application.id) && !app_p.rejected(@application.id)%>
```
juliet-e, line 16: I like how you've organized this rendering on this view, but I wonder if you could get these if statements down to 2 conditions rather than 3, for the sake of developer empathy. Maybe a model method in the application_pets model that checks both whether an app_pet is approved OR rejected?

app/views/shelters/edit.html.erb
```diff
@@ -1,16 +1,5 @@
...
-
-  <%= f.label :foster_program %>
-  <%= f.check_box :foster_program %>
-
-  <%= f.label :rank %>
-  <%= f.number_field :rank %>
-
-  <%= f.submit %>
-<% end %>
+<%= render partial: 'form', locals: {
```
juliet-e, line 2: Nice use of a partial!

config/routes.rb
```diff
@@ -36,4 +46,10 @@
   get "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinary_offices#veterinarians"
   get "/veterinary_offices/:veterinary_office_id/veterinarians/new", to: "veterinarians#new"
   post "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinarians#create"
-end
+
+  get "/admin/shelters", to: "admin/shelters#index"
+  get "/admin/applications", to: "admin/applications#index"
+  get "/admin/applications/:id", to: "admin/applications#show"
+  patch "/admin/applications/:id", to: "admin/application_pets#update"
```
juliet-e, line 53: Referring back to the comment I left in your admin/application_pets_controller -- this is not an external route that a user has to go to, but is a route for a data operation to update the application_pets resource. So, it's okay to just put it in the other ApplicationPetsController, and then after updating, you can redirect back to an admin page

db/seeds.rb
```diff
@@ -1,7 +1,38 @@
-# This file should contain all the record creation needed to seed the database with its default values.
-# The data can then be loaded with the bin/rails db:seed command (or created alongside the database with db:setup).
-#
-# Examples:
-#
-#   movies = Movie.create([{ name: "Star Wars" }, { name: "Lord of the Rings" }])
-#   Character.create(name: "Luke", movie: movies.first)
+Pet.destroy_all
```
juliet-e, line 1: It's  hard to say what the error was that you got from Render when dealing with seeds, but my guess is that you deleted pets that were associated with an ApplicationPet and Application record(s). To avoid this error, you could first include ApplicationPet.destroy_all, and then proceed with these other deletes. That's because your ApplicationPet table has foreign keys to other tables so those need to be deleted first before the Pet or Application record is deleted. 

Another option is to check out `dependent: :destroy`

app/views/admin/shelters/index.html.erb
```diff
@@ -0,0 +1,16 @@
+<h2>All Shelters</h2>
+
+<% @shelters.each do |shelter| %>
+  <div id="shelter-<%= shelter.id %>">
```
juliet-e, line 4: This formatting looks set up well to use within blocks. I wonder if the issue you had arose from a syntax issue. The syntax can be tricky!! I would save this line to remember what the syntax can look like. What trips me up a lot is that the erb tags here have to be render tags (i.e. you need the `=`) even though the div tag is not going to render on the page. The ruby needs to return a value to populate the div id tag.

spec/models/pet_spec.rb
```diff
@@ -39,4 +39,18 @@
...
+  end
+
+  it "has an application approval verification for the pet" do
+    @shelter_1 = Shelter.create(name: "Aurora shelter", city: "Aurora, CO", foster_program: false, rank: 9)
+    @pet_1 = Pet.create(adoptable: true, age: 7, breed: "sphynx", name: "Bare-y Manilow", shelter_id: @shelter_1.id)
+    @application_1 = Application.create!(name: "Stacy Chapman", street_address: "1870 Canopy Rd", city: "Los Angeles", state: "CA", zip_code: 90001, description: "I grew up with dachshunds and felt really connected", status: "In Progress")
+    @app_pet_1 = ApplicationPet.create!(application_id: @application_1.id, pet_id: @pet_1.id)
+    @app_pet_1.approve
+    expect(@pet_1.application_approved?).to eq(true)
+  end
```
juliet-e, line 55: Nice work testing these model methods!

spec/spec_helper.rb
```diff
@@ -92,3 +92,9 @@
   Kernel.srand config.seed
 =end
 end
+
+RSpec.configure do |config|
+  config.before(:suite) do
+    Rails.application.load_seed # loading seeds
```
juliet-e, line 98: This is the reason your tests were failing - you had duplicates in the test database because you were running seeds AND setting up test data. You should do one or the other. At the beginning, we encourage folks to set up their own test data in the tests rather than just using seeds so that you can have more control and see what data exists in each test. You also don't need much test data in lots of the early tests. If you want to use seeds for test data, that's okay, just in that case, you shouldn't also create additional data in a before each because that will create additional data!

PR #151, opened 2023-10-26
Blaine/Edward

juliet-e:
Creating this PR to leave you feedback! :)


app/controllers/admin/application_pets_controller.rb
```diff
@@ -0,0 +1,12 @@
+class Admin::ApplicationPetsController < ApplicationController
```
juliet-e, line 1: Nice work using a joins controller! This is an "internal" path though, in that a user doesn't need to write the URL in the navbar, but a request arrives here from a button. So, there's no real need for this to have `/admin` at the beginning just cause it's not a path a user is going to. This just means that you can put this `update` in the regular ApplicationPetsController. This isn't wrong by any means but it's okay to group it with the other joins actions if you want to

app/controllers/admin/application_pets_controller.rb
```diff
@@ -0,0 +1,12 @@
...
+  def update
+    @application_pet = ApplicationPet.where("pet_id = ? AND application_id = ?", params[:pet_id], params[:application_id])
```
juliet-e, line 3: I'd suggest `find_by` here and pass in the pet_id and application_id because it will return just one record rather than an array!

app/controllers/admin/shelters_controller.rb
```diff
@@ -0,0 +1,10 @@
+class Admin::SheltersController < ApplicationController
+  def index
+    @shelters = Shelter.reverse_abc
```
juliet-e, line 3: beautiful encapsulation of business logic into model methods! These controllers are so nice and clean

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+      description: params[:description],
+      pets: [],
+      status: "In Progress"
+    })
+
+    if @application.save
+      redirect_to "/applications/#{@application.id}"
+    else
+      flash[:alert] = "Error: Fill in all fields"
+      redirect_to "/applications/new"
```
juliet-e, line 34: Great inclusion of sad path handling

app/models/application.rb
```diff
@@ -0,0 +1,18 @@
...
+  validates :city, presence: true
+  validates :state, presence: true
+  validates :zip_code, presence: true
+  validates :description, presence: true
+  validates :status, presence: :true, inclusion: {in: ["In Progress", "Accepted", "Rejected", "Pending"]}
+
+  def search_pets_by_name(pet_name)
+    return [] unless pet_name.present?
+
+    Pet.where("name ILIKE ?", "%#{pet_name}%")
```
juliet-e, line 16: This is a brownfield project, so there's some code already written that you should leverage! Check out the `.search` in the ApplicationRecord file. That's the parent class for all model classes, so you could do `Pet.search` and it would delegate to that parent method! No need to write the query that already exists, just test it

app/models/application_record.rb
```diff
@@ -1,6 +1,6 @@
 class ApplicationRecord < ActiveRecord::Base
   primary_abstract_class
-  
+
   def self.search(search_params)
```
juliet-e, line 4: Here it is! You can use this directly

app/controllers/applications_controller.rb
juliet-e: You can also render :new instead of redirect

app/models/application_record.rb
juliet-e: Even when you use a parent class' method, it's good to test it for this new use case (i.e. searching for pets)

app/models/shelter.rb
```diff
@@ -4,6 +4,7 @@ class Shelter < ApplicationRecord
   validates :city, presence: true
 
   has_many :pets, dependent: :destroy
+  has_many :applications, through: :pets
```
juliet-e, line 7: Great work adding this additional association! You can see when you do `Shelter.joins(:applications)` that SQL is under the hood, using the joins table and pets table to get to this one, but this AR association adds a nice shortcut to make the syntax much easier!

app/views/admin/applications/_button.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= form_with url: url, method: method do |f| %>
```
juliet-e, line 1: This is an interesting design choice because we already have the `button_to` helper, so I'm curious what trade offs you see between this partial you made, and the button_to that comes with Rails?

app/views/admin/applications/show.html.erb
```diff
@@ -0,0 +1,52 @@
...
+</ul>
+</section>
+
+<h2>Pets Applied for</h2>
+<ul>
+  <% @application.pets.each do |pet| %>
+ <br><%= link_to "#{pet.name}", "/pets/#{pet.id}" %>
+  <% if pet.application_pet(@application.id).status.nil? %>
+        <%= render partial: 'button', locals: {
+                                      url: "/admin/application_pets/#{@application.id}?pet_id=#{pet.id}&application_id=#{@application.id}&status=approved",
```
juliet-e, line 32: You have your application_id twice here -- once as a path variable and another as a query parameter. You can just use one of these! You could also feel free to remove the `/admin` to the beginning since this is sort of an "internal" route, not one that the user sees. Another option is adding the pet ID directly to the path rather than a query parameter. like so: "/applications/:application_id/pets/:pet_id?status=approved" ---> still a RESTful path!

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,62 @@
...
+    <% end %>
+  <% end %>
+</section>
+<br>
+
+<section id="pet-search">
+<% if @application.status == "In Progress" %>
+  <h3>Add a Pet to this Application</h3>
+  <%= form_with url: "/applications/#{@application.id}", method: :get, data: { turbo: false } do |form| %>
+    <%= form.hidden_field :search, value: "pet" %>
```
juliet-e, line 44: you could probably get rid of this hidden field, and just check in the controller action for the presence of `pet_name` in the params object

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,62 @@
...
+  <% end %>
+<% end %>
+</section>
+
+<br>
+
+<section id="adopt-a-pet">
+<% if !@search_pets.nil? && @search_pets.any? %>
+  <% @search_pets.each do |pet| %>
+    <%= form_with url: "/application_pets?pet_id=#{pet.id}&application_id=#{@application.id}", method: :post, data: { turbo: false } do |form| %>
```
juliet-e, line 57: Similar idea here -- this is a RESTful route, but if you wanted to, you could also put the IDs in the path and it would still be considered RESTful as long as you're mentioning both applications and pets in the path to indicate the association between these resources. 
example: "/applications/:application_id/pets/:pet_id"

app/views/pets/new.html.erb
```diff
@@ -1,16 +1,5 @@
...
-
-  <%= f.label :age %>
-  <%= f.number_field :age %>
-
-  <%= f.label :adoptable %>
-  <%= f.check_box :adoptable, value: true %>
-
-  <%= f.submit %>
-<% end %>
+<%= render partial: "form", locals: {
```
juliet-e, line 2: Nice use of partials!

config/routes.rb
```diff
@@ -1,6 +1,14 @@
 Rails.application.routes.draw do
   get "/", to: "application#welcome"
 
+  get "/applications", to: "applications#index"
+  get "/applications/new", to: "applications#new"
+  get "/applications/:id", to: "applications#show"
+  post "/applications", to: "applications#create"
+  patch "/applications/:id", to: "applications#update"
+
+  post "/application_pets", to: "application_pets#create"
```
juliet-e, line 10: Great RESTful routing!

db/seeds.rb
```diff
@@ -5,3 +5,18 @@
 #
 #   movies = Movie.create([{ name: "Star Wars" }, { name: "Lord of the Rings" }])
 #   Character.create(name: "Luke", movie: movies.first)
+
+# Shelter.destroy_all
+# Pet.destroy_all
+# Application.destroy_all
```
juliet-e, line 11: You'll need to destroy your seeds in order to prevent duplicates, just make sure you're careful about the order in which you're destroying them (or check out `dependent: :destroy`

For example, you can't destroy Shelters before you destroy Pets because Pets have a foreign key to a shelter. So after you run line 9, there are pets in the database with invalid foreign keys to shelters. So you need to delete any tables with foreign keys (think child tables or joins tables) **before** you delete any parent tables.

pull_request_template.md
```diff
@@ -0,0 +1,20 @@
...
+- [ ] docs
+
+## Checklist
+
+- [ ] code has been self reviewed
+- [ ] code runs without any errors
+- [ ] thorough testing has been implemented if adding feature
+- [ ] all tests pass
+
+### Thanks!
```
juliet-e, line 20: Love to see a PR template!

spec/models/application_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    it { should validate_presence_of(:street_address) }
+    it { should validate_presence_of(:city) }
+    it { should validate_presence_of(:state) }
+    it { should validate_presence_of(:zip_code) }
+    it { should validate_presence_of(:description) }
+    it { should validate_presence_of(:status) }
+  end
+
+  describe ".search_pets_by_name(pet_name)" do 
+    it "should return pets that have been searched" do
```
juliet-e, line 20: Nice work testing this model method! You can  use this same test and just change your implementation to use the already made .search

spec/models/shelter_spec.rb
```diff
@@ -67,5 +68,18 @@
...
+    describe ".reverse_abc" do
+      it "returns the shelters in reverse order" do
+        expect(Shelter.reverse_abc).to eq([@shelter_2, @shelter_3, @shelter_1])  
+      end
+    end
+    
+    describe ".pending_applications" do
+      it "returns the pending applications" do 
+       
+        expect(Shelter.pending_applications).to eq([])
```
juliet-e, line 81: This is not the most robust test because the test data provided isn't extensive enough. This doesn't verify that the query you have is returning the correct thing since it's returning an empty array. It should probably return the @application you have in your test data, no? Or you can create another test application and see what the output is

app/models/shelter.rb
```diff
@@ -31,4 +32,15 @@ def alphabetical_pets
...
+
+  def self.reverse_abc
+    find_by_sql("
+      SELECT * FROM shelters
+      ORDER BY name DESC
+    ")
+  end
+
+  def self.pending_applications
+    Shelter.joins(pets: :applications).where("applications.status = ?", "Pending").order(:name)
```
juliet-e, line 44: since you have this shortcut, you can actually do `Shelter.joins(:applications)`

spec/models/shelter_spec.rb
juliet-e: You'll also need to create some ApplicationPets in order to this application to be associated to a shelter

PR #152, opened 2023-10-26
Eric & Michael Adopt Don't Shop Eval

dcoleman21:


app/controllers/shelters_controller.rb
```diff
@@ -4,6 +4,10 @@ def index
       @shelters = Shelter.order_by_number_of_pets
     elsif params[:search].present?
       @shelters = Shelter.search(params[:search])
+    elsif params[:admin] == "admin"
+      @admin = true
+      @shelters = Shelter.order_by_sql
+      @shelters_with_applications = Shelter.pending_applications
```
dcoleman21, line 10: We don't want to create more things in the controller then we need. Remember we want our controllers to be as slim as possible. Line 10 can go away and you can just call `.pending_applications` on your instance `@shelters` in your view.

app/controllers/shelters_controller.rb
```diff
@@ -22,7 +26,17 @@ def pets
...
+      @admin = true
+      @current_shelter = Shelter.find(params[:id])
+      @shelter = Shelter.name_and_address(params[:id])
+      @pending_pets = Shelter.find(params[:id]).pending_pets
+      @age_stats = Shelter.age_stats(@current_shelter)
+      @pet_count = Shelter.pet_count(@current_shelter)
+      @pets_with_homes = Shelter.pets_with_homes(@current_shelter)
+    else
+      @shelter = Shelter.find(params[:id])
+    end
```
dcoleman21, line 39: There is way too much going on in here. We realize that the extensions are doing a lot but this is a `show` action so we are just looking for a single instance of a shelter. So line 31 is finding the shelter, now in your view you can call all of these methods on that one instance vs creating all of these additional instance variables in the controller. 
Also these methods (`name_and_address`, `pending_pets`, `age_stats`, `pet_count`, etc) should not be class methods but instance methods. So you would need to change that in your `shelter` model.

app/models/application.rb
```diff
@@ -0,0 +1,54 @@
...
+      app_pets
+    end
+  end
+
+  def find_rejections
+    pets = self.pets.map { |pet| pet.id }
+    pets.reduce([]) do |rej_pets, pet|
+      rej_pets << pet if ApplicationPet.where("pet_id = '#{pet}' and application_id = '#{self.id}' and rejected = true") != []
+      rej_pets
+    end
```
dcoleman21, line 52: These two methods have a lot of Ruby going on.

app/models/shelter.rb
```diff
@@ -16,6 +16,39 @@ def self.order_by_number_of_pets
...
+    end.count
+  end
+
+  def self.pets_with_homes(shelter)
+    shelter.pets.flat_map do |pet|
+      pet.applications.select do |application|
+        application.status == "Approved"
+      end
+    end.count
+  end
```
dcoleman21, line 50: The methods on lines 34 - 50 should not be class methods but instance methods.

app/views/shelters/_form.html.erb
```diff
@@ -0,0 +1,15 @@
...
+  <%= f.text_field :city %>
+
+  <%= f.label :foster_program %>
+  <%= f.check_box :foster_program %>
+
+  <%= f.label :rank %>
+  <%= f.number_field :rank %>
+
+  <%= f.submit %>
+<% end %>
```
dcoleman21, line 15: Nice job building a partial

app/views/shelters/edit.html.erb
```diff
@@ -1,16 +1,6 @@
 <h1>Edit Shelter</h1>
-<%= form_with url: "/shelters/#{@shelter.id}", method: :patch, local: true do |f| %>
-  <%= f.label :name %>
-  <%= f.text_field :name %>
+<%= render partial: "form", locals: {
+  path: "/shelters/#{@shelter.id}",
+  method: :patch
+} %>
```
dcoleman21, line 5: nice job using your partial

app/views/shelters/show.html.erb
```diff
@@ -1,8 +1,33 @@
...
+<% if @admin %>
 
-<p class="pet-count">Pets at <%= @shelter.name %>: <%= @shelter.pet_count %></p>
-<p><%= link_to "All pets at #{@shelter.name}", "/shelters/#{@shelter.id}/pets" %></p>
+  <h3><%= @shelter %></h3>
+ 
 
-<p><%= link_to "Delete #{@shelter.name}", "/shelters/#{@shelter.id}", method: :delete %></p>
+  <h4> <%= "Action Required" %> </h4>
+  <% @pending_pets.each do |pet| %>
```
dcoleman21, line 7: Here is where I would just have the `@current_shelter` instance variable in your controller and then in this view you could call `@current_shelter.pending_pets` because you should be able to reach through the shelter to get to its pets by the associations that you have.

config/routes.rb
```diff
@@ -36,4 +36,15 @@
...
+  get "/applications/new", to: "applications#new"
+  post "/applications", to: "applications#create"
+  get "/applications/:id", to: "applications#show"
+  patch "/applications/:id", to: "applications#update"
+  
+  get "/:admin/applications/:id", to: "applications#show"
+  patch "/:admin/applications/:app_id/:pet_id/:app_rej", to: "application_pets#update"
+  
+  get "/:admin/shelters", to: "shelters#index"
+  get "/:admin/shelters/:id", to: "shelters#show"
```
dcoleman21, line 49: I think Juliet talked to you about this during a check in, we would want to use name spacing for all admin functionality. So you would create a directory called admin and then have a file called shelters. 
```ruby
controllers/admin/shelters_controller.rb
controllers/admin/applications_controller.rb
```

config/routes.rb
```diff
@@ -36,4 +36,15 @@
...
   post "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinarians#create"
+
+  get "/applications/new", to: "applications#new"
+  post "/applications", to: "applications#create"
+  get "/applications/:id", to: "applications#show"
+  patch "/applications/:id", to: "applications#update"
+  
+  get "/:admin/applications/:id", to: "applications#show"
+  patch "/:admin/applications/:app_id/:pet_id/:app_rej", to: "application_pets#update"
+
```
dcoleman21, line 47: a more RESTful route for this would be
```ruby
patch "/applications/:application_id/pets/:pet_id", to: "application_pets#update"
```
Let your view handle the if/else logic for `accept/reject`.

spec/features/shelters/index_spec.rb
```diff
@@ -105,4 +106,86 @@
     expect(page).to have_content(@shelter_2.name)
     expect(page).to_not have_content(@shelter_1.name)
   end
+
+  it "Admin page will list shelter in reverse alphabetical order" do
+    visit "/admin/shelters"
+
+    expect("RGV animal shelter").to appear_before("Fancy pets of Colorado")
+    expect("Fancy pets of Colorado").to appear_before("Aurora shelter")
+  end
```
dcoleman21, line 115: Nice job using the `orderly gem`

spec/features/shelters/index_spec.rb
```diff
@@ -105,4 +106,86 @@
...
+    application2.pets << @pet_3 
+
+    visit "/admin/shelters"
+
+    within("div#admin") do
+      expect(page).to have_content("Shelters with Pending Applications")
+      expect(page).to have_content(@shelter_1.name)
+      expect(page).to_not have_content("Fancy pets of Colorado")
+      expect(page).to_not have_content("RGV animal shelter")
+    end
```
dcoleman21, line 133: Nice job using `within` blocks

spec/models/application_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+    it "Checks if an application has_pets?" do
+      application = Application.create!(name: "Eric", full_address: "9999 Street Road, Denver, CO 80231", good_home: "Gimme", good_owner: "I like cats", status: "Pending")
+      expect(application.has_pets?).to be false
+      application.pets << @pet_1
+      expect(application.has_pets?).to be true
+    end
+  end
+
+  describe '#Application creation with missing params' do
+    before :each do
```
dcoleman21, line 54: These tests are failing because you cannot create an application without a name - your validations prevent this from happening.

spec/models/application_spec.rb
```diff
@@ -0,0 +1,122 @@
...
+      # Will give a link and error message to which field is empty
+      params = {name: "", full_address: "123 Test St", good_home: "I like fish"}
+      expect(Application.empty_params_link(params)).to eq("/applications/new?name=true")
+      # Will redirect to the page that you'ree supposed to go to without errors
+      params = {name: "Eric", full_address: "123 Test St", good_home: "I like fish"}
+      expect(Application.empty_params_link(params)).to eq("/applications/new")
+      # give multiple errors in the link for which fields are empty.
+      params = {name: "", full_address: "", good_home: ""}
+      expect(Application.empty_params_link(params)).to eq("/applications/new?name=true&full_address=true&good_home=true")
+    end
```
dcoleman21, line 76: What are you trying to test here?

dcoleman21 (commented):
Overall nice job. I would continue to work on improving your overall MVC implementations. We want slim controllers (meaning we only want what we need) so don't build a whole bunch of instance variables when we can reach through our objects in the views and call our model methods on those instances. Let the views handle more of the if/else logic. Let the models perform the database logic/querying using AR/SQL vs ruby. Some ruby is okay but we should be able to do most with AR/SQL. We also need to understand better when to make methods class vs instance. Also we want to build RESTful routes. So when we are creating new relationships between two existing resources (ie pets and applications) we want our routes to give us the params that we need and point to the join controller with CRUD actions. Let us know if you have any questions.

PR #153, opened 2023-10-26
Kam/Justin

juliet-e:
Creating this to leave you feedback on your project!

### DIFF
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -4,5 +4,13 @@ class AdminController < ApplicationController
     @pendingshelters = Shelter.shelters_apps_in_progress
   end
 
+  def show
+    @app = App.find(params[:id])
+  end
 
+  def update
+    @app = App.find(params[:id])
+    @app.update!(status: 2)
+    redirect_to "/admin/apps/#{params[:id]}"
+  end
 end
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -10,7 +10,11 @@ class AdminController < ApplicationController
 
   def update
     @app = App.find(params[:id])
-    @app.update!(status: 2)
+    if params[:approved] == "true"
+      @app.update!(status: 2)
+    elsif params[:rejected] == "true"
+      @app.update!(status: 3)
+    end
     redirect_to "/admin/apps/#{params[:id]}"
   end
 end
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -1,5 +1,8 @@
 class AdminController < ApplicationController
   def index
     @shelters = Shelter.order_by_name_reverse
+    @pendingshelters = Shelter.shelters_apps_in_progress
   end
+
+
 end
--- a/app/controllers/admin_controller.rb
+++ b/app/controllers/admin_controller.rb
@@ -1,0 +1,5 @@
+class AdminController < ApplicationController
+  def index
+    @shelters = Shelter.order_by_name_reverse
+  end
+end
--- a/app/controllers/apps_controller.rb
+++ b/app/controllers/apps_controller.rb
@@ -1,9 +1,23 @@
 class AppsController < ApplicationController
   def index
-
+    @apps = App.all
+  end
+  
+  def new
+    @app = App.new
   end
 
   def show
     @apps = App.find(params[:id])
+  end
+
+  def create
+    app = App.create(app_params)
+    redirect_to "/apps/#{app.id}"
+  end
+
+  private
+  def app_params
+    params.permit(:name, :address, :city, :zip, :description)
   end
 end
--- a/app/controllers/apps_controller.rb
+++ b/app/controllers/apps_controller.rb
@@ -9,7 +9,6 @@ class AppsController < ApplicationController
   end
 
   def show
-    
     @app = App.find(params[:id])
     # require 'pry'; binding.pry
     if params[:search].present?
@@ -19,7 +18,12 @@ class AppsController < ApplicationController
 
   def create
     app = App.create(app_params)
-    redirect_to "/apps/#{app.id}"
+    if app.save
+      redirect_to "/apps/#{app.id}"
+    else
+      redirect_to "/apps/"
+      flash[:alert] = "Error: Please fill out this field}"
+    end
   end
 
   def update
--- a/app/controllers/apps_controller.rb
+++ b/app/controllers/apps_controller.rb
@@ -9,6 +9,7 @@ class AppsController < ApplicationController
   end
 
   def show
+    
     @app = App.find(params[:id])
     # require 'pry'; binding.pry
     if params[:search].present?
@@ -22,12 +23,14 @@ class AppsController < ApplicationController
   end
 
   def update
-    # require 'pry'; binding.pry
     @app = App.find(params[:id])
-    @pet = Pet.find(params[:pet_id])
-    # require 'pry'; binding.pry
-    @app.pets << @pet
-    redirect_to "/apps/#{params[:id]}"
+    if !params[:good_owner].present?
+      @pet = Pet.find(params[:pet_id])
+      @app.pets << @pet
+    else 
+      @app.update(status: 1)
+    end
+      redirect_to "/apps/#{params[:id]}"
   end
 
   def add_pet
@@ -41,6 +44,6 @@ class AppsController < ApplicationController
   
   private
   def app_params
-    params.permit(:name, :address, :city, :zip, :description, status: "In Progress")
+    params.permit(:name, :address, :city, :zip, :description)
   end
 end
--- a/app/controllers/apps_controller.rb
+++ b/app/controllers/apps_controller.rb
@@ -1,0 +1,9 @@
+class AppsController < ApplicationController
+  def index
+
+  end
+
+  def show
+    @apps = App.find(params[:id])
+  end
+end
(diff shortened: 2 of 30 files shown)


app/controllers/admin_controller.rb
```diff
@@ -0,0 +1,20 @@
+class AdminController < ApplicationController
```
juliet-e, line 1: Since there might be *lots* of admin paths, it's generally more conventional to do one of the following:
1. Create an admin folder under controllers, and then under the admin folder, make whatever controllers named after the resources they're dealing with - i.e. admin/shelters_controller, admin/applications_controller etc. WHen you set it up this way, your class declaration becomes `class Admin::SheltersController` to indicate that it lives in the admin folder
2. name the controllers with admin in the name: admin_shelters_controller, admin_applications_controller

In both of these cases, controllers are still based off the data they're manipulating (i.e. shelters or applications) but not just "admin".

app/controllers/admin_controller.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  def show
+    @app = App.find(params[:id])
+  end
+
+  def update
+    @app = App.find(params[:id])
+    if params[:approved] == "true"
+      @app.update!(status: 2)
```
juliet-e, line 14: Great work using enums for status!

juliet-e: For readability, you can use the string or symbol version of the enum instead of the integer and Rails is smart enough to map it to the integer when it gets to the database

app/controllers/apps_controller.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  def create
+    app = App.new(app_params)
+    if app.save
+      redirect_to "/apps/#{app.id}"
+    else
+      flash[:alert] = "Error: Please fill out all fields"
+      render :new
+    end
```
juliet-e, line 24: Great sad path handling!

app/models/admin.rb
```diff
@@ -0,0 +1,6 @@
+class Admin < ApplicationRecord
+  
+  def self.order_by_name_reverse
```
juliet-e, line 3: Admin doesn't map to a database table, so it shouldn't have a model. Instead, it's just being used as a namespace, or a tool to organize routes. This query could live in the Shelters model and be tested in the Shelters_spec

app/models/pet.rb
```diff
@@ -10,4 +12,8 @@ def shelter_name
   def self.adoptable
     where(adoptable: true)
   end
+
+  def self.search(search_params)
```
juliet-e, line 16: This is a brownfield project, so you should look around at the existing code and see what code you could leverage for your needs! In this case, there's an already written search method in the ApplicationRecord parent class, which is inherited by every model. That means that you can call any model class with `.search` and get a search output. Your query looks great, but just wanted to point out that this code is already written for you!

app/models/shelter.rb
```diff
@@ -16,6 +16,12 @@ def self.order_by_number_of_pets
       .order("pets_count DESC")
   end
 
+  def self.shelters_apps_in_progress
+    pets = Pet.find(AppsPet.pluck(:pet_id))
+    shelter_ids = pets.pluck(:shelter_id)
+    shelters = Shelter.where(id: shelter_ids)
```
juliet-e, line 22: Does this query actually verify that a shelter has an app that is "Pending"? Consider something like this: `Shelter.joins(pets: :applications).where("applications.status = 'Pending'")`

app/models/shelter.rb
```diff
@@ -31,4 +37,8 @@ def alphabetical_pets
   def shelter_pets_filtered_by_age(age_filter)
     adoptable_pets.where("age >= ?", age_filter)
   end
+
+  def self.order_by_name_reverse
```
juliet-e, line 41: Nice! This is where this shelter query should live (you can probably delete the admin model you have)

app/views/admin/show.html.erb
```diff
@@ -0,0 +1,25 @@
...
+  <%= link_to "#{pet.name}", "/pets/#{pet.id}" %> 
+  
+  <%if @app.status == "Accepted" %>
+    <p>your application for <%=pet.name %> has been Approved!</p>
+
+  <%elsif @app.status == "Rejected"%>
+    <p>your application for <%=pet.name %> has been Rejected!</p>
+
+  <% else @app.status == "Pending" %>
+    <%= button_to "approve application to adopt #{pet.name}", "/admin/apps/#{@app.id}?approved=true", method: :patch%></p>
```
juliet-e, line 17: A more RESTful route here would include pets in the path to indicate that you are updating the joins record between applications and pets. Consider something like "/applications/:application_id/pets/:pet_id"

juliet-e: and then you can still have a query parameter for "approved" or "rejected"!

app/views/apps/new.html.erb
```diff
@@ -0,0 +1,6 @@
+<%= render partial: 'form', locals: {
```
juliet-e, line 1: Nice use of a partial!

app/views/apps/show.html.erb
```diff
@@ -0,0 +1,49 @@
...
+
+  <% if params[:pet_name].present? %>
+    <h3>Pets matching search</h3>
+    <% @pets = Pet.search(params[:pet_name]) %>
+    <% @pets.each do |pet| %>
+      <p><%= link_to "#{pet.name}", "/pets/#{pet.id}" %></p>
+      <% if @app.pets.include?(pet) %>
+        <p>This pet is already on this application</p>
+      <% else %>
+        <%= button_to "Adopt #{pet.name}", "/apps/#{@app.id}?app_id=#{params[:id]}&pet_id=#{pet.id}&commit=Add+Pet", method: :patch %>
```
juliet-e, line 35: The RESTful path here would also include both applications and pets, and could be a POST because what you are doing is creating an association between pets and applications. We want to reserve a patch to the applications controller for updating attributes on an application table. But here, we are creating a joins record. So consider this path: POST to "applications/:app_id/pets"

app/views/apps/show.html.erb
```diff
@@ -0,0 +1,49 @@
...
+      <% end %>
+    <% end %>
+  <% end %>
+
+<% if @app.pets.count > 0 && @app.status == "In Progress"%>
+<h3>Submit my application</h3>
+  <%= form_with url: "/apps/#{@app.id}?commit=Submit+Application", method: :patch, local: true do |form|%>
```
juliet-e, line 42: By making the above change, this also means you don't need this update action to distinguish between submitting application and adding a pet!

config/routes.rb
```diff
@@ -36,4 +36,16 @@
   get "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinary_offices#veterinarians"
   get "/veterinary_offices/:veterinary_office_id/veterinarians/new", to: "veterinarians#new"
   post "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinarians#create"
+
+  get "/apps", to: "apps#index"
+  get "apps/new", to: "apps#new"
+  get "/apps/search", to: "apps#show"
+  get "apps/:id", to: "apps#show"
+  post "/apps", to: "apps#create"
+  patch "/apps/:id", to: "apps#update"
```
juliet-e, line 45: See notes in view and controllers about more RESTful approaches to these routes! These are things we're all still learning so it's very normal to get this type of feedback on this project!

db/seeds.rb
```diff
@@ -5,3 +5,21 @@
 #
 #   movies = Movie.create([{ name: "Star Wars" }, { name: "Lord of the Rings" }])
 #   Character.create(name: "Luke", movie: movies.first)
+ Shelter.destroy_all
```
juliet-e, line 8: Make sure you check out `dependent: :destroy` or only delete the child objects (like pets) or join object (application_pets) first before destroying the parent objects. Otherwise, you will get errors when running seeds locally and when deploying!

spec/models/admin_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+      @app1 = App.create(name: "Timmy", address: "123 Main St", city: "Aurora", state: "CO", zip: 80111, description: "I love dogs")
+      @pet2 = @shelter1.pets.create(adoptable: true, age: 3, breed: "doberman", name: "Lobster")
+      @pet3 = @shelter2.pets.create(adoptable: true, age: 2, breed: "saint bernard", name: "Beethoven")
+    end
+
+    it "lists all the shelters in the system in reverse alphabetical order" do
+      expect(Shelter.order_by_name_reverse).to eq([@shelter2, @shelter3, @shelter1])
+    end
+  end
+end
```
juliet-e, line 19: This should be tested in the shelter_spec and this admin model actually doesn't need to be here! You can safely delete this test and admin.rb

app/views/apps/_form.html.erb
```diff
@@ -0,0 +1,16 @@
...
+  <p><%= form.label :address %>
+  <%= form.text_field :address, placeholder: "ex. 123 Drewry lane"%></p>
+  <p><%= form.label :city %>
+  <%= form.text_field :city, placeholder: "ex. Denver"%></p>
+  <p><%= form.label :state %>
+  <%= form.text_field :state, placeholder: "ex. CO"%></p>
+  <p><%= form.label :zip %>
+  <%= form.text_field :zip, placeholder: "ex. 12345"%></p>
+  <p><%= form.label :description %>
+  <%= form.text_field :description, placeholder: "Why are you good home?"%></p>
```
juliet-e, line 13: re: your question in your video: You could pass a local variable that's a boolean (like `show_description`) and then use an if statement with that boolean in the partial to decide whether to include that field. If there's a lot of variability with what fields should should up, though, I probably wouldn't use a partial

PR #155, opened 2023-10-28
Matt/Victor ADS Eval

juliet-e:


app/controllers/admin/shelters_controller.rb
```diff
@@ -0,0 +1,5 @@
+class Admin::SheltersController < ApplicationController
+  def index
+    @shelters = Shelter.order_reverse_name
```
juliet-e, line 3: Great encapsulation of this data logic into a model method! :)

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def create
+    application = Application.new(application_params)
+    if application.save
+      redirect_to "/applications/#{application.id}"
+    else
+      flash[:alert] = "Please fill in all required fields"
+      # flash[:alert] = "Error: #{error_message(@application.errors)}"
+      redirect_to "/applications/new"
+      # render :new
```
juliet-e, line 14: Great work handling the sad path! You can either redirect OR render the form again here. The only difference is that `render` will not make an additional HTTP request and will just render the form again. This means, the form could end up on the page again without changing the URI from the previous request

app/controllers/pet_applications_controller.rb
```diff
@@ -0,0 +1,12 @@
+class PetApplicationsController < ApplicationController
```
juliet-e, line 1: Excellent controller organization!

app/controllers/pet_applications_controller.rb
```diff
@@ -0,0 +1,12 @@
...
+  def create
+    pet_application = PetApplication.new({application_id: params[:application_id], pet_id: params[:pet_id]})
+    if pet_application.save
+      redirect_to "/applications/#{params[:application_id]}"
+    else
+      redirect_to "/applications/#{params[:application_id]}"
+      flash[:alert] = "Error: #{error_message(application.errors)}"
+    end
+  end
+end
```
juliet-e, line 12: For US 11/12, you would probably eventually end up changing the status of a PetApplication record based on when an individual pet is approved. This logic could be done in this controller in an `update` action

app/models/application.rb
```diff
@@ -0,0 +1,14 @@
+class Application < ApplicationRecord
+  has_many :pet_applications
+  has_many :pets, through: :pet_applications
+  validates :name, :street_address, :city, :state, :zip_code, :description, presence: true
+  #enum application_status: {"In Progress": 0, "Pending": 1, "Accepted":2, "Rejected":3}
+  
+  def in_progress?
+    application_status == "In Progress"
+    #use an enum vs a method at the same level as the has_many
```
juliet-e, line 9: This is a great option when your field is a string datatype! If your field is an enum type, you get this method created out of the box. For example, with an enum, every enum value gets turned automatically into a boolean method that will return true if the field contains that value. So you would be able to call `application.accepted?` or `application.pending?` without having to implement those methods because they are made for you when using an enum

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,47 @@
...
+<% unless @application.pets.empty? %>
+    <h2>Pets You've Applied For:</h2>
+      <% @application.pets.each do |pet| %>
+        <p><%= link_to pet.name, "/pets/#{pet.id}" %></p>
+      <% end %>
+<% end %>
+
+<% if @application.application_status == "In Progress" %>
+    <h1>Add a pet to this application</h1>
+    <%= form_with url: "/applications/#{@application.id}", method: :get, local: true do |form| %>
```
juliet-e, line 19: I encourage you to replace `local: true` with `data: {turbo: false}` because we're using Rails 7 and Turbo can sometimes interfere with form requests made that redirect back to the same controller action (which this page does!)

juliet-e: I see now that you're disabling Turbo below and I'm wondering if you saw different issues with this form that made you bring back `local: true`? Let me know!

config/routes.rb
```diff
@@ -1,6 +1,10 @@
 Rails.application.routes.draw do
   get "/", to: "application#welcome"
 
+  get "/admin/shelters", to: "admin/shelters#index"
+
+  post "/pet_applications", to: "pet_applications#create"
```
juliet-e, line 6: Excellent RESTful routing! Another RESTful route here can include the IDs like so if you don't want to deal with sending hidden values: `/applications/:app_id/pets/:pet_id`
Either path is fine and RESTful!

app/controllers/admin/shelters_controller.rb
```diff
@@ -0,0 +1,5 @@
+class Admin::SheltersController < ApplicationController
```
juliet-e, line 1: Nice work namespacing this controller under admin!

spec/features/applications/show_spec.rb
```diff
@@ -0,0 +1,197 @@
...
+
+      visit "/applications/#{application.id}"
+
+      fill_in(:pet_name, with: "Lucille Bald")
+      click_button("Search for Pets")
+      expect("Lucille Bald").to_not appear_before("Search for Pets")
+      expect(page).to have_button("Adopt this Pet")
+      click_button("Adopt this Pet")
+      expect(current_path).to eq("/applications/#{application.id}")
+      expect("Lucille Bald").to appear_before("Search for Pets")
```
juliet-e, line 82: Another option here in order to make sure that Lucille Bald is in the right section of page is to use **within blocks**. Add a div or a section to your view with an ID called `associated-pets` or something, and then you can do:
```
within "#associated-pets" do
    expect(page).to have_content("Lucille Bald")
end
```

app/models/pet.rb
```diff
@@ -10,4 +12,8 @@ def shelter_name
   def self.adoptable
     where(adoptable: true)
   end
+
+  def get_pet_by_name(name)
+    Pet.where(name: name)
```
juliet-e, line 17: Remember that every model method you write should be tested in a model spec! You could also do `find_by(name: xxx)` and do this in the controller

app/models/shelter.rb
```diff
@@ -16,6 +16,10 @@ def self.order_by_number_of_pets
       .order("pets_count DESC")
   end
 
+  def self.order_reverse_name
+    Shelter.find_by_sql("SELECT * FROM shelters ORDER BY name desc")
```
juliet-e, line 20: Don't forget to model test this!!

PR #156, opened 2023-10-28
Joseph/Shawn ADS Eval

juliet-e:


app/controllers/admin/applications_controller.rb
```diff
@@ -0,0 +1,6 @@
+class Admin::ApplicationsController < ApplicationController
```
juliet-e, line 1: Great controller organization using the Admin namespace!

app/controllers/application_pets_controller.rb
```diff
@@ -0,0 +1,27 @@
+class ApplicationPetsController < ApplicationController
+  def update
+    application = Application.find(params[:id])
```
juliet-e, line 3: This is very minor, but from a routing perspective, I would assume the ID in `/application_pets/:id` would be the primary key for that ApplicationPet record rather than the application ID. So, an equally RESTful route with the IDs of the separate resources (applications and pets) could be: `"/applications/:application_id/pets/:pet_id"`

You can also keep "/application_pets/:id"  as it is RESTful, but it would be more intuitive to pass the joins record ID here (but it is not as easily readable from your view, hence why it's a bit tricker for this context)

app/controllers/application_pets_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+    selected_app_pet = ApplicationPet.find(params[:pet_id])
+    if params[:approved?] == "Yes"
```
juliet-e, line 5: Refactor idea to consider: Your if statement could just check for the presence of a `:approved` param with `params[:approved].present?` and then it could pass the value of the param down to a helper method or model method to handle updating the AP status and adoptability based on the value of that param. That could clean up this controller action a bit and consolidate logic

Also, just to avoid issues with URL syntax, query parameters shouldn't include a `?` because that is the path symbol to indicate the beginning of query parameters. We generally only add `?` to boolean methods.

app/controllers/application_pets_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+      selected_app_pet.update(adoptable: false)
+    else
+      selected_app_pet.update(application_pet_status: 3)
+      selected_app_pet.pet.update(adoptable: true)
+    end
+    update_application(application)
+    redirect_to "/admin/applications/#{application.id}"
+  end
+
+  def update_application(application)
```
juliet-e, line 16: Nice delegation to a helper method! I think this could potentially live in the Application model to allow for more easily unit testing it

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,38 @@
...
+      @application.update(status: 1)
+    end
+    redirect_to "/applications/#{@application.id}"
+  end
+
+  private
+  def application_params
+    params.permit(:name, :street_address, :city, :state, :zip_code, :description, :status)
+  end
+end
```
juliet-e, line 38: Excellent RESTfulness and adherence to MVC here!

app/helpers/states_helper.rb
```diff
@@ -0,0 +1,59 @@
+module StatesHelper
```
juliet-e, line 1: Love the additional validation you've included!!!

app/models/application.rb
```diff
@@ -0,0 +1,24 @@
...
+  validates :name, :street_address, :city, :state, :description, presence: true
+  enum status: {"In Progress": 0, "Pending": 1, "Accepted": 2, "Rejected": 3}
+  validates :zip_code, presence: true, numericality: true, length: {minimum: 5, maximum: 5}
+
+  def full_address
+    "#{self.street_address}, #{self.city}, #{self.state} #{self.zip_code}"
+  end
+
+  def add_pet(pet)
+    ApplicationPet.create(application: self, pet: pet, name: pet.name, application_pet_status: 1, adoptable: true)
```
juliet-e, line 14: It's totally okay to encapsulate this, but it is also okay to do this in the controller! This is a simple CRUD action so it's not an MVC violation to create the record in the controller action

app/models/pet.rb
```diff
@@ -10,4 +12,8 @@ def shelter_name
   def self.adoptable
     where(adoptable: true)
   end
+
+  def self.search_for_pet(search)
+    where('lower(name) like ?', "%#{search.downcase}%")
```
juliet-e, line 17: This search method is actually already implemented for you in the ApplicationRecord! This is a bit of an Easter Egg (although you're not penalized for not finding it!). Because this is a brownfield project, we want you to explore the codebase and leverage the methods already created (like you did with  the `error_messages` helper method in ApplicationController). So, you could've used this `.search` method from the model parent class and just tested it in order to complete this user story

app/models/shelter.rb
```diff
@@ -4,6 +4,7 @@ class Shelter < ApplicationRecord
   validates :city, presence: true
 
   has_many :pets, dependent: :destroy
+  has_many :applications, through: :pets
```
juliet-e, line 7: Nice work adding this additional AR association!

app/views/admin/applications/show.html.erb
```diff
@@ -0,0 +1,64 @@
...
+<div class="body" >
+<p><u>Address</u>: <%= @application.full_address %> </p>
+<p><u>Why would you make a good home?</u>: <%= @application.description %> </p>
+<p><u>Status</u>: <%= @application.status %> </p>
+
+<p><u>Pets I'd Like to Adopt</u>:</p>
+<% @application.application_pets.each do |pet| %>
+<div class= "buttons">
+  <% if pet.application_pet_status == "Pending" && pet.adoptable == true %>
+    <%= form_with url:"/application_pets/#{@application.id}", method: :patch, local: :true do |f| %>
```
juliet-e, line 25: See note in controller about this path -- it's expected that the ID you pass here is the application_pet record ID so this path might've been a better choice (but both are RESTful): "applications/:app_id/pets/:pet_id"

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,54 @@
...
+<p><u>Status</u>: <%= @application.status %> </p>
+
+<p><u>Pets I'd Like to Adopt</u>:</p>
+<% @pets.each do |pet| %>
+  <%= link_to "#{pet.name}", "/pets/#{pet.id}", method: :get %>
+<%end%>
+
+<% if @application.status == "In Progress" %>
+  <h2><u>Add a pet to this application</u>:</h2>
+  <%= form_with url:"/applications/#{@application.id}", method: :get, local: :true do |f| %>
```
juliet-e, line 27: Sometimes Turbo gets mad when we create forms that redirect back to the same controller action -- if you saw any weirdness with that, make sure you include `data: { turbo: false }`

config/routes.rb
```diff
@@ -1,6 +1,16 @@
 Rails.application.routes.draw do
   get "/", to: "application#welcome"
 
+  get "/admin/applications/:application_id", to: "admin/applications#show"
```
juliet-e, line 4: Excellent RESTful routing

spec/models/shelter_spec.rb
```diff
@@ -67,5 +97,29 @@
...
+      it "returns number of adopted pets in shelter" do
+        expect(@shelter_1.adopted_pets_total).to eq(1)
+      end
+    end
+
+    describe ".action_required" do
+      it "returns shelter pets with pending applications" do
+        expect(@shelter_1.action_required).to eq(["Mr. Pirate"])
+      end
+    end
```
juliet-e, line 123: Excellent attention to unit and end-to-end testing!

app/models/pet.rb
juliet-e: Actually, it looks like your tests are testing the `.search` function that's already created instead of this method, so you can remove this method and make sure you're calling .search from your controller!

PR #157, opened 2023-10-28
Blake/Sooyung Adopt Don't Shop Eval

juliet-e:
Creating this to leave you project feedback!


config/routes.rb
```diff
@@ -36,4 +36,15 @@
   get "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinary_offices#veterinarians"
   get "/veterinary_offices/:veterinary_office_id/veterinarians/new", to: "veterinarians#new"
   post "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinarians#create"
+
+  post "/applications", to: "applications#create"
+  get "/applications/new", to: "applications#new"
+  get "/applications/:id", to: "applications#show", as: 'show_application'
+  post "/applications/:id", to: "pet_applications#create", as: 'create_pet_application'
```
juliet-e, line 43: When we are creating an association between 2 existing resources like we are here, a more RESTful option for the path would be:
POST to "/applications/:application_id/pets/:pet_id" or "application_pets"
Your controller action is very RESTful as you are creating the record in a `create` action in a joins controller, but the path could be more RESTful

config/routes.rb
```diff
@@ -36,4 +36,15 @@
...
+  patch "/applications/:id", to: "applications#update"
+
+  get "/admin/shelters", to: "admin/shelters#index"
+
+  get "/admin/applications/:id", to: "admin/applications#show"
+  patch "/admin/applications/:id", to: "admin/applications#update"
```
juliet-e, line 49: This action seems to be updating the pet_application rather than the pet, so a better RESTful choice here would be
`patch "/applications/:app_id/pets/:pet_id", to: "pet_applications#update"`
Note that we don't need the `admin` prepended here because this is not an external route that a user goes to, but a route that is triggered from clicking a button on that page. Then, once updating the pet application, you can redirect to an admin path

app/controllers/admin/applications_controller.rb
```diff
@@ -0,0 +1,17 @@
+class Admin::ApplicationsController < ApplicationController
+  def show
+    @application = Application.find(params[:id])
+  end
+
+  def update
+    application = Application.find(params[:id])
+    pet_app = PetApplication.find_by(pet_id: params[:pet_id])
```
juliet-e, line 8: You can send more than one param to find_by, so you could do `PetApplication.find_by(pet_id: params[:pet_id], application_id: params[:application_id])`

app/controllers/admin/applications_controller.rb
```diff
@@ -0,0 +1,17 @@
+class Admin::ApplicationsController < ApplicationController
+  def show
+    @application = Application.find(params[:id])
+  end
+
+  def update
```
juliet-e, line 6: See note in routes about this action -- this update could live in your PetApplicationsController

app/controllers/pet_applications_controller.rb
```diff
@@ -0,0 +1,12 @@
+class PetApplicationsController < ApplicationController
```
juliet-e, line 1: Excellent work creating this Joins controller! A very RESTful approach to setting up many to many joins records

app/controllers/pet_applications_controller.rb
```diff
@@ -0,0 +1,12 @@
...
+  def create 
+    @application = Application.find(params[:id])
+    @pet = Pet.find(params[:format]) 
+
+    unless @application.pets.include?(@pet)
```
juliet-e, line 6: Nice edge case handling here

app/models/pet.rb
```diff
@@ -10,4 +12,8 @@ def shelter_name
   def self.adoptable
     where(adoptable: true)
   end
+
+  def self.adoptable_search(search_params)
```
juliet-e, line 16: This search method is actually already implemented for you in the ApplicationRecord! This is a bit of an Easter Egg (although you're not penalized for not finding it!). Because this is a brownfield project, we want you to explore the codebase and leverage the methods already created (like you did with  the `error_messages` helper method in ApplicationController). So, you could've used this `.search` method from the model parent class and just tested it in order to complete this user story

app/models/shelter.rb
```diff
@@ -31,4 +31,12 @@ def alphabetical_pets
   def shelter_pets_filtered_by_age(age_filter)
     adoptable_pets.where("age >= ?", age_filter)
   end
+
+  def self.reverse_alphabetical  
+    find_by_sql("SELECT * FROM shelters ORDER BY name DESC")
+  end
+
+  def self.pending_apps
+    joins(pets: :applications).where("applications.status = '1'").distinct
```
juliet-e, line 40: This looks great! Another option here that I'm sharing for fun (and to help with the next project!)
You can set up additional AR associations in order to make this join a bit easier. I call it "mooching off the associations of your children". Pets has associations with applications, and pets belong to shelters. So, shelters can take advantage of pets' associations. If you add the following to this Shelter model:
`has_many :applications, through: :pets`
then you can do this:
`Shelter.joins(:applications)` directly!

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,46 @@
...
+
+<h3>Pets Applied for: </h3>
+<% @application.pets.each do |pet| %>
+<p><%= link_to "#{pet.name}", "/pets/#{pet.id}" %> </p>
+<% end %> 
+<h3>Application Status: </h3> <%= @application.status %>
+
+<% if @application.status == "In Progress" %>
+  <h3> Add a Pet to this Application:</h3>
+  <%= form_with url: "/applications/#{@application.id}?pet_name=#{params[:pet_name]}&commit=Search", method: :get do %>
```
juliet-e, line 15: If you're using a form, you shouldn't need to have the query parameters in this path (i.e. you shouldn't need to include `?pet_name=#{params[:pet_name]}&commit=Search` in the url on line 15 as they should be added automatically. Any field you put in a form will be put in the params object!

juliet-e: Also, sometimes Turbo complains about forms that redirect back to the same path, so if you saw any weirdness, remember to do `data: {turbo: false}`!

spec/features/admin/applications/show_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+        application1.pets << pet_1
+        application1.pets << pet_3
+
+        visit "/applications/#{application1.id}"
+        fill_in :description, with: "I know how to take care pets!"
+        click_button "Submit Application"
+
+        visit "/admin/applications/#{application1.id}"
+
+        within("tr:contains('#{pet_1.name}')") do
```
juliet-e, line 40: Nice use of within blocks!!

spec/features/applications/show_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+      expect(page).to have_content("Pom Pom")
+    end
+    
+    it 'adds the pet to under Pets Applied for' do
+      #US 5
+      application1 = Application.create!(name: "Hannah Banana", street_address: "1234 Sugarwood Cir", city: "Newport", state: "Kentucky", zip_code: "41071", description: "I already have a cat and my cat Dave needs a friend. Dave is very friendly and other cat would be a great addition for our household!")
+      visit "/applications/#{application1.id}"
+      fill_in "Enter pet name", with: "Pom Pom"
+      click_button "Search"
+      click_button "Adopt this Pet"
```
juliet-e, line 85: It looks like your before_each already adds Pom Pom to this application. This test should add a pet that doesn't already have an association to the application (like maybe pet 2?)

spec/helpers/welcome_helper_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+#
+# describe WelcomeHelper do
+#   describe "string concat" do
+#     it "concats two strings with spaces" do
+#       expect(helper.concat_strings("this","that")).to eq("this that")
+#     end
+#   end
+# end
+RSpec.describe WelcomeHelper, type: :helper do
+  pending "add some examples to (or delete) #{__FILE__}"
```
juliet-e, line 14: Please don't use any scaffolding! This looks like it was generated by a Rails scaffolding command

spec/models/shelter_spec.rb
```diff
@@ -67,5 +73,17 @@
...
+      it "returns shelters in reverse alphabetical order" do
+        expect(Shelter.reverse_alphabetical).to eq([@shelter_2, @shelter_3, @shelter_1])
+      end
+    end
+
+    describe "pending application shelters" do 
+      it "returns only shelters with applications that are pending" do 
+        expect(Shelter.pending_apps).to eq([@shelter_1])
+      end
+    end
```
juliet-e, line 87: Nice work testing at the unit (model spec) and end-to-end level (feature spec!)

PR #158, opened 2023-10-28
Brendan/Kevin ADS eval

juliet-e:


app/controllers/admin_applications_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+  end
+  
+  def show
+    @application = Application.find(params[:id])
+  end
+
+  def update
+    pet = Pet.find(params[:pet_id])
+    application = Application.find(params[:id])
+    pet_application = PetApplication.where("pet_id = #{pet.id}", "application_id = #{application.id}")
```
juliet-e, line 13: Per your question in your video: a slightly more MVC-friendly method here could be `find_by`. You could do: `pet_application = PetApplication.find_by(pet_id = pet.id, application_id = application.id)`
It's not wildly different than a where, but because where clauses are often use for more complex querying (i.e. things beyond exact value searches) we encourage students to remember `.where` as a model method clause. IMO this is not a huge MVC violation but I would swap it out with find_by and use key/value syntax to do exact value searching.

app/controllers/admin_applications_controller.rb
```diff
@@ -0,0 +1,18 @@
+class AdminApplicationsController < ApplicationController
```
juliet-e, line 1: For this project, this controller organization is great. IT is possible, however, that this controller could get VERY big if all the admin paths lead here. So another option is to use an admin *directory* in your controller directory, and then create 2 different controllers under the admin directory -> "/admin/shelters_controller.rb" and "admin/applications_controller.rb". In each of those controller files, the class declaration would look like this: `class Admin::ApplicationsController < ApplicationController` with the `Admin::` namespace to indicate that this controller lives within the admin directory.

We'll talk more about this concept on Tuesday in class!

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+      city: params[:city],
+      state: params[:state],
+      zipcode: params[:zipcode],
+      description: params[:description],
+      status: "In Progress"
+    })
+    if @application.valid?
+      redirect_to "/applications/#{@application.id}"
+    else 
+      @application.save
```
juliet-e, line 30: Instead of doing `.valid` and `.save` in two different steps, you can do it at once and just use `.save` (which will run .valid under the hood)
```
@application = Application.new(params...)
if @application.save
    happy_path_stuff
else
   flash errors by accessing `@application.errors.full_messages`
end

File: app/controllers/applications_controller.rb [hunk 4]
```diff
@@ -0,0 +1,44 @@
...
+    if params[:pet_id].present?
+      @application.add_pet(params[:pet_id])  
+    end
+  end
+
+  def new
+  end
+
+  def create
+    admin = Admin.create
```
juliet-e, line 17: `admin` in our path is actually not a model, but just an organizational tool. We can call it a namespace. It shouldn't actually map to a model or database table. So you can keep applications as a standalone object rather than something that belongs to an admin!

config/routes.rb
```diff
@@ -29,11 +29,22 @@
   patch "/veterinarians/:id", to: "veterinarians#update"
   delete "/veterinarians/:id", to: "veterinarians#destroy"
 
-  get "/shelters/:shelter_id/pets", to: "shelters#pets"
+  get "/shelters/:shelter_id/pets", to: "shelter_pets#index"
```
juliet-e, line 32: Love this change! Refactoring legacy code to make it more RESTful 🦾

app/controllers/shelter_pets_controller.rb
```diff
@@ -0,0 +1,13 @@
+class ShelterPetsController < ApplicationController
```
juliet-e, line 1: YESSS!! Nice refactor to make a more RESTful controller set up for this legacy code!

app/models/application.rb
```diff
@@ -0,0 +1,27 @@
...
+  validates :name, presence: true
+  validates :address, presence: true
+  validates :city, presence: true
+  validates :state, presence: true
+  validates :zipcode, presence: true
+  validates :description, presence: true
+
+  def add_pet(pet_id)
+    pet = Pet.find(pet_id)
+    pets << pet
```
juliet-e, line 14: Nice encapsulation of this logic! This is not strictly necessary from an MVC perspective but I do think it cleans up this functionality into a nice modular action

app/models/shelter.rb
```diff
@@ -4,6 +4,9 @@ class Shelter < ApplicationRecord
   validates :city, presence: true
 
   has_many :pets, dependent: :destroy
+  has_many :applications, through: :pets
```
juliet-e, line 7: Nice work setting up this additional association! Adding this makes the syntax for joining below so much simpler. Mooching off of child model's associations is always a great choice!

app/models/admin.rb
```diff
@@ -0,0 +1,4 @@
+class Admin < ApplicationRecord
```
juliet-e, line 1: No need for this model! Admin is a namespace rather than a resource. We will talk more about this on Tuesday! The general rule of thumb -- if there's something in a path that doesn't map to a resource in the database or a tangible piece of data, it is likely an organizational tool and can be used as a resource.

app/models/admin_shelters.rb
```diff
@@ -0,0 +1,4 @@
+class AdminShelters < ApplicationRecord
```
juliet-e, line 1: No need for this model either. You can grab applications from shelters by adding the extra association that you have in your shelter model that builds a shortcut from Shelters to Applications

app/views/admin_applications/show.html.erb
```diff
@@ -0,0 +1,18 @@
...
+
+<% @application.pet_applications.each do |pet_application| %>
+  <section id="pet<%= pet_application.pet_id %>">
+    <%= pet_application.pet.name %>
+    <% if pet_application.approved %>
+      </br>This pet is approved
+    <% elsif pet_application.rejected %>
+      </br>This pet is rejected
+    <% else %>
+      <%= button_to "Approve this Pet", "/admin/applications/#{@application.id}", method: :patch, params: { approved: true, pet_id: pet_application.pet_id }, data: { turbo: false } %>
```
juliet-e, line 14: What we're doing here is actually updating the joins record that is associating 2 existing resources: a pet and an application. So, it's best to have our route reflect these 2 resources. The 2 most common RESTful paths include:
"/applications/:app_id/pets/:pet_id" or "application_pets/:id" where this last ID is the primary key for the joins table record associating these 2

Note that neither of these have `admin` prepended to them, because this route is triggered from a button rather than a URI that is entered by the user, so it's sort of an internal route rather than external. We could definitely keep `admin` but it's not strictly necessary. So it could route to `pet_applications#update`

juliet-e: Also, Turbo shouldn't interfere with buttons so you should be able to remove the `data: { turbo: false }`

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,38 @@
...
+        <%= f.text_field :search_app %>
+        <%= f.submit "Submit" %>
+        <% end %>
+    <% end %></br>
+</section>
+
+<% if @pets.present? %>
+    <% @pets.each do |pet| %>
+        <%= pet.name %> 
+        <%= button_to "Adopt this Pet", "/applications/#{@application.id}", method: :get, params: { pet_id: pet.id } %>
```
juliet-e, line 25: when we click this button, should it be a GET request? Keep in mind we are *creating* a joins record between a pet and an application in this case, so we could probably use a POST here instead. Similar to the previous comment around paths for joins records, a more RESTful route here could be: "/applications/:app_id/pets" to indicate you are joining these 2 resources

app/views/veterinarians/_form.html.erb
```diff
@@ -0,0 +1,12 @@
+<%= form_with url: path, method: method, local: true, data: { turbo: false } do |f| %>
```
juliet-e, line 1: Y'all did a lot of DRYing up your forms all over the application! Nice work

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,44 @@
+class ApplicationsController < ApplicationController
+  def show
+    @application = Application.find(params[:id])
+    if params[:search_app].present?
+      @pets = Pet.search(params[:search_app])
+    end
+
+    if params[:pet_id].present?
```
juliet-e, line 8: This controller action is doing a bit too much. `show` actions are for reading data so it shoudn't handle creating the join record between a pet and application. Instead, you can have that button to Adopt a Pet route to a joins controller, called ApplicationPetsController or PetApplicationsController. You could route it to a `create` action there to indicate the creation of a joins resource.

juliet-e: Then, you can route the app back to the show page

config/routes.rb
```diff
@@ -29,11 +29,22 @@
...
   get "/shelters/:shelter_id/pets/new", to: "pets#new"
   post "/shelters/:shelter_id/pets", to: "pets#create"
 
-  get "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinary_offices#veterinarians"
+  get "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinary_office_veterinarians#index"
   get "/veterinary_offices/:veterinary_office_id/veterinarians/new", to: "veterinarians#new"
   post "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinarians#create"
+
+  get "/applications/new", to: "applications#new"
+  get "/applications/:id", to: "applications#show"
```
juliet-e, line 41: See note in applications_controller about this routing. For adding a pet to an app, you could route the button to:
`post "/applications/:id/pets", to: "pet_applications#create"`

Similarly, when you approve an application, you could route that approval button to
`patch "/applications/:app_id/pets/:pet_id", to: "pet_applications#update"`

spec/features/applications/new_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    visit "/applications/new"
+    fill_in "name", with: "Bob"
+    fill_in "address", with: "123 1st St"
+    fill_in "city", with: "Pleasantville"
+    fill_in "state", with: "CO"
+    fill_in "zipcode", with: "80501"
+
+    click_button "Submit"
+    
+    expect(page).to have_current_path("/applications")
```
juliet-e, line 41: This test could be a bit clearer about expecting a flash error message. If your logic for displaying flash messages wasn't working, check out [this](https://backend.turing.edu/module2/lessons/sad_path_and_flash) lesson plan to set up flash display in the application layout

spec/models/shelter_spec.rb
```diff
@@ -13,56 +15,78 @@
...
       end
     end
 
-    describe "#order_by_number_of_pets" do
+    describe ".order_by_number_of_pets" do
       it "orders the shelters by number of pets they have, descending" do
         expect(Shelter.order_by_number_of_pets).to eq([@shelter_1, @shelter_3, @shelter_2])
       end
     end
+
```
juliet-e, line 56: Nice unit testing of this model logic!

PR #183, opened 2023-12-14
Meg and Yain - Project Submission

megroth2:


app/models/application.rb
```diff
@@ -0,0 +1,41 @@
...
+  validates :street_address, presence: true
+  validates :city, presence: true
+  validates :state, presence: true
+  validates :zipcode, presence: true, numericality: true
+  validates :description, presence: true
+  has_many :application_pets
+
+
+  def self.application(id)
+    Application.find(id)
```
juliet-e, line 12: This is a bit of encapsulation overkill IMO -- you can just call .find from the controller and that's not a violation of MVC!

app/controllers/admins/shelters_controller.rb
```diff
@@ -0,0 +1,6 @@
+class Admins::SheltersController < ApplicationController
+  def index
+    @shelters = Shelter.order_by_reverse_alphabetically
+    @shelters_pending = Shelter.pending_applications
```
juliet-e, line 4: Great encapsulation of these data queries!

app/controllers/admins/shelters_controller.rb
```diff
@@ -0,0 +1,6 @@
+class Admins::SheltersController < ApplicationController
```
juliet-e, line 1: Excellent work using the admin namespace

app/controllers/application_pets_controller.rb
```diff
@@ -0,0 +1,23 @@
+class ApplicationPetsController < ApplicationController
```
juliet-e, line 1: Nice use of the joins controller!

app/controllers/application_pets_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  def create
+  @application_pet = ApplicationPet.create(application_pet_params)
+
+    redirect_to show_application_path
+  end
+
+  def update
```
juliet-e, line 9: Excellent adherence to RESTful conventions with these controller actions in the joins controller

app/models/application.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def full_address
+    street_address << " " << city << ", " << state << " " << zipcode
+  end
+
+  def list_of_pets
+    Pet.joins(:applications).where("application_id = ?", self.id)
```
juliet-e, line 20: This query is equivalent to `self.pets` -- it's a bit easier/more readable just to use the associations rather than use a join

app/models/application.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def added_pets?
+    list_of_pets.present?
+  end
+
+  def all_pets_approved
+    if status_of_application_pet.uniq.count == 1 && status_of_application_pet.first
```
juliet-e, line 28: The following solution would work best if you had a pet_application status that had a default value like "unreviewed" rather than `nil` but you could do something like:

```
non_approved_pets = self.application_pets.where("application_approved != 'Approved'")
return true if non_approved_pets.empty? 
false
```

This includes more AR and less Ruby

app/models/application_pet.rb
```diff
@@ -0,0 +1,7 @@
+class ApplicationPet < ApplicationRecord
+  validates :application_id, presence: true
+  validates :pet_id, presence: true
+
+  belongs_to :application
+  belongs_to :pet
```
juliet-e, line 6: Great many to many setup!

app/models/pet.rb
```diff
@@ -2,12 +2,26 @@ class Pet < ApplicationRecord
...
   belongs_to :shelter
+  has_many :application_pets
+  has_many :applications, through: :application_pets
 
   def shelter_name
     shelter.name
   end
 
+  def self.find_pet(id)
+    find(id)
```
juliet-e, line 13: Same idea here -- no need to encapsulate this in a model method, it's very conventional to keep it in the controller!

app/models/shelter.rb
```diff
@@ -16,6 +16,15 @@ def self.order_by_number_of_pets
...
 
+  def self.order_by_reverse_alphabetically
+    Shelter.find_by_sql("select * from shelters order by name desc")
+  end
+
+  def self.pending_applications
+    Shelter.joins(pets: :applications).where("applications.status = ?", "Pending").order(:name).distinct
+  end
+
+
```
juliet-e, line 27: Excellent SQL and AR querying here

app/views/admins/applications/show.html.erb
```diff
@@ -0,0 +1,30 @@
...
+  <h1>Pet Application For <%= @application.name %> Is Approved </h1>
+<% elsif @application.all_pets_approved == nil %>
+  <h1>Pending Pet Application For <%= @application.name %> </h1>
+  <p>Pets applied for:</p>
+  <ul>
+    <% @application.list_of_pets.each do |pet| %>
+    <div id="pet-<%= pet.id %>">
+      <li><%= pet.name %></li>
+        <ul>
+          <% if pet.application_pet_approved(params[:id]) %>
```
juliet-e, line 11: There's some complex display logic here and you're not going to get rid of all of it (you will definitely have if statements here) but one option for consolidating some of this logic would be to display the status of the PetApplication dynamically. You could check if a PetApplication has a status, and if it does, just print the value with a label like so:
`<p>Status: <%= pet.application_pet.status %> </p>`

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,59 @@
...
+  </section>
+  <% end %>
+
+<% elsif @application.status == "Pending" %>
+
+  <p>Why I would make a good owner: <%= @application.good_owner_comments %></p>
+  
+<% else %>
+
+  <section id="add-pet">
```
juliet-e, line 43: This section should be able to be consolidated so that it's not duplicated in the view. The sections of the page could be restructured like so (pseudocode below):
```
if status is not pending
    - option to search for pets
    - if pets are already present on the app
              -area to display associated pets
              -option to submit
else # i.e. status is pending
     show why applicant would make a good owner



```

app/views/pets/_form.html.erb
```diff
@@ -0,0 +1,15 @@
+<%= form_with url: path, method: method, data: { turbo: false } do |f| %>
```
juliet-e, line 1: Nice inclusion of partial!

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,59 @@
+<h1><%= @application.name %>'s Application</h1>
+<p>Address: <%= @application.full_address %></p>
```
juliet-e, line 2: Could these application details be encapsulated into a partial for use in the admin show page as well?

spec/features/admins/applications/show_spec.rb
```diff
@@ -0,0 +1,168 @@
...
+    @application_pet_1 = ApplicationPet.create!(application_id: @application_2.id, pet_id: @pet_1.id)
+    @application_pet_2 = ApplicationPet.create!(application_id: @application_3.id, pet_id: @pet_3.id)
+    @application_pet_2 = ApplicationPet.create!(application_id: @application_2.id, pet_id: @pet_3.id)
+    @application_pet_2 = ApplicationPet.create!(application_id: @application_3.id, pet_id: @pet_4.id)
+  end
+
+  it "has a button to approve an application for every pet that the application pet is for" do
+    # User Story 12
+    visit "/admin/applications/#{@application_2.id}"
+    within "#pet-#{@pet_1.id}" do
```
juliet-e, line 32: great use of within blocks to narrow down the scope of the page!

spec/features/applications/show_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+      expect(page).to have_content("Add a Pet to this Application")
+      expect(page).to have_content("Search for Pets by name:")
+      expect(page).to have_button("Submit")
+      expect(page).to have_no_content("Hamster")
+      expect(page).to have_no_button("Adopt this pet")
+
+      fill_in("Search for Pets by name:", with: "Hamster")
+      click_button("Submit")
+
+      expect(page).to have_content("Add a Pet to this Application")
```
juliet-e, line 53: Once you've tested for the presence of this text above, I don't think you need to verify it's still there, especially since you're verifying the current path of the applicationon line 59

spec/features/applications/show_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+      expect(page).to have_content("Search for Pets by name:")
+      expect(page).to have_button("Submit")
+      expect(page).to have_content("Hamster")
+      expect(page).to have_button("Adopt this pet")
+
+      expect(page.current_path).to eq("/applications/#{@application_1.id}")
+      expect(page).to have_content("Hamster")
+    end
+
+    it "will let me add a pet to my application" do
```
juliet-e, line 63: This test and the one before it could be consolidated to avoid duplicate actions/assertions

spec/features/applications/show_spec.rb
```diff
@@ -0,0 +1,176 @@
...
+      expect(page).to have_no_content("Submit Application")
+    end
+
+    it "will not show the submit section if I have not added any pets" do
+      # User Story 7
+      visit "/applications/#{@application_with_no_pets.id}"
+
+      expect(page).to have_no_content("Why I would make a good owner")
+      expect(page).to have_no_button("Submit Application")
+    end
```
juliet-e, line 174: Excellent robust testing!

spec/models/pet_spec.rb
```diff
@@ -39,4 +49,16 @@
...
     end
   end
+
+  describe "#self.find_pet(id)" do
+    it "finds the pet with a matching ID" do
+      expect(Pet.find_pet(@dog.id)).to eq(@dog)
+    end
+  end
+
+  describe "#adoptable?" do
```
juliet-e, line 59: Did you mean for this test to be for #already_approved?

spec/models/shelter_spec.rb
```diff
@@ -41,31 +56,56 @@
...
       end
     end
-
-    describe ".pet_count" do
-      it "returns the number of pets at the given shelter" do
-        expect(@shelter_1.pet_count).to eq(3)
-      end
-    end
   end
+
```
juliet-e, line 110: Your testing overall is excellent! At both the unit (model spec) and integration (feature spec) level, you're capturing reasonable sad paths and watching for logical edge cases. Nice work!

PR #184, opened 2023-12-14
Matt Darlington & Isaac Mitchell

mdarl17:


app/controllers/admin/applications_controller.rb
```diff
@@ -0,0 +1,6 @@
+class Admin::ApplicationsController < ApplicationController
```
juliet-e, line 1: Awesome use of the admin namespace to organize your controllers!

app/controllers/admin/shelters_controller.rb
```diff
@@ -0,0 +1,13 @@
+class Admin::SheltersController < ApplicationController 
+  def index 
+    @shelters = Shelter.all
+  end
+
+  def show
+    @shelter = Shelter.find(params[:id])
+    #wondering which method to put these ^^ in for admin.
```
juliet-e, line 8: There's no need to create an additional admin/shelter model as this isn't a new database, but rather a different page to view the same shelter data. `admin` is a namespace, which is really just a tool for organizing routes (in this case, distinguishing the applicant-facing routes from the admin-facing routes) rather than relating to a specific database resource. We'll learn more about namespacing in the Advanced Routing class after break.

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,45 @@
+class ApplicationsController < ApplicationController
+  def show
+    @application = Application.find(params[:id])
+    if params[:pet_name].present?
+      @pets = Pet.search(params[:pet_name])
```
juliet-e, line 5: Nice work using the existing `search` method from ApplicationRecord!

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+      city: params[:city],
+      state: params[:state],
+      zip: params[:zip],
+      descr: params[:good_home]
+    })
+
+    if application.save
+      redirect_to "/applications/#{application.id}"
+    else 
+      flash[:alert] = "Application not created: Required information missing."
```
juliet-e, line 28: Nice sad path handling. Keep in mind in the future that you can use dynamic messaging for these errors by sending the specific AR error in the flash: `flash[:alert] = application.errors.full_messages`

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+
+      redirect_to "/applications/new"
+    end
+  end
+
+  def update 
+    @application = Application.find(params[:id])
+    @pets = @application.pets
```
juliet-e, line 36: Is this @pets instance var necessary?

app/controllers/pet_applications_controller.rb
```diff
@@ -0,0 +1,30 @@
+class PetApplicationsController < ApplicationController
```
juliet-e, line 1: Nice use of the joins controller to adhere to RESTful routing!! Using `create` for adding a pet to an application and `update` to setting the status of the pet_application is the most RESTful choice

app/models/admin/shelter.rb
```diff
@@ -0,0 +1,14 @@
+class Admin::Shelter < ApplicationRecord
+  validates :name, presence: true
+  validates :rank, presence: true, numericality: true
+  validates :city, presence: true
+
+  has_many :pets, dependent: :destroy
+  
+  def find_shelter_name(shelter_id)
```
juliet-e, line 8: As mentioned in the controller, it's not necessary to add an additional Admin::Shelter model but to just build this query in the existing Shelter model. When we add the the admin/shelter page, we are not adding additional data but just seeing the data displayed differently

app/models/application.rb
```diff
@@ -0,0 +1,26 @@
+class Application < ApplicationRecord
+  has_many :pet_applications
+  has_many :pets, through: :pet_applications
+ 
+  validates_presence_of :name, :street, :city, :state, :zip, :descr, :status
+
+  enum status: {"In progress": 0, "Pending": 1, "Approved": 2, "Rejected": 3}
```
juliet-e, line 7: Nice use of enums!

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,45 @@
+class ApplicationsController < ApplicationController
+  def show
+    @application = Application.find(params[:id])
+    if params[:pet_name].present?
+      @pets = Pet.search(params[:pet_name])
+    else
+      @pets = @application.pet_search(params[:pet_name])
```
juliet-e, line 7: I'm curious what led you to this solution with `pet_search`. It seems it will only be reached with the `params[:pet_name]` is not present, which also means there are no search terms available yet. So, in the view, you can display the seach results only if @pets is available. I don't quite understand why you are populating @pets in this case when the search term isn't present

app/models/application.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def pet_search(name)
+    Pet.where("name = ?", name)
```
juliet-e, line 10: Again, not sure what the purpose of this method is - is it necessary?

app/models/application.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def get_pet_app(pet_id) 
+    PetApplication.where(pet_id: pet_id, application_id: self.id).first
```
juliet-e, line 14: Another option for the future is `find_by` which can take more than one attribute for searching and only returns one record rather than an array

app/models/application.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def all_pets_apps_appr?
+    if self.pet_applications.where('status = 2').count == self.pet_applications.count &&
```
juliet-e, line 18: I wonder if you could simplify this logic by instead using `where("status != 2")` to find any cases where there are pet_apps that aren't approved. The elsif statement here is not as readable but I understand what you're accomplishing here!

app/models/shelter.rb
```diff
@@ -16,6 +16,29 @@ def self.order_by_number_of_pets
...
+      "SELECT shelters.*
+      FROM shelters 
+      ORDER BY shelters.name DESC"
+    )
+  end
+
+  def self.pending_apps 
+    select("shelters.*")
+      .distinct
+      .joins(pets: { pet_applications: :application })
```
juliet-e, line 30: Nice work! These queries look great. Another option, since our connection between pets and applications is set up through associations, is to join them directly like so`.joins(pets: :applications` (i.e. you don't need to mention pet_applications since you have a `has_many x through pet_applications` in the model already)

app/models/shelter.rb
```diff
@@ -16,6 +16,29 @@ def self.order_by_number_of_pets
...
+      .where("applications.status = ?", 1)
+  end
+
+  def get_name 
+    Shelter.find_by_sql("SELECT shelters.name FROM shelters WHERE id = #{self.id}").first.name
```
juliet-e, line 35: self.name would work here too

app/models/shelter.rb
```diff
@@ -16,6 +16,29 @@ def self.order_by_number_of_pets
...
+  end
+
+  def get_address
+    Shelter.find_by_sql("SELECT shelters.city FROM shelters WHERE id = #{self.id}").first.city
```
juliet-e, line 39: Similarly: self.city

app/views/admin/applications/show.html.erb
```diff
@@ -0,0 +1,15 @@
+<h1>Admin Applications</h1> 
+<% @application.pets.each do |pet| %>
+  <%= form_with url: "/pet_applications/#{@application.get_pet_app(pet.id).id}?app_id=#{@application.id}&pet_id=#{pet.id}", method: :patch, local: true do |f|  %>
```
juliet-e, line 3: This is a nice RESTful route as it names both resources involved in the join. However it can be a pain to have to add all those query params so another equally RESTful option is: "/applications/:app_id/pets/:pet_id"

app/views/admin/shelters/show.html.erb
```diff
@@ -0,0 +1,2 @@
+<h1> <%= @shelter.get_name %> </h1>
```
juliet-e, line 1: @shelter.name and @shelter.city would be okay here too

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,64 @@
...
+    <%= form_with url: "/applications/#{@application.id}", method: :get, local: true do |form| %>
+      <%= form.hidden_field :search_pet  %>
+      <%= form.label "Search for pet" %>
+      <%= form.text_field :pet_name%>
+      <%= form.submit "Search" %>
+    <% end %>
+  
+    <% @pets.each do |pet| %> 
+    <section id=<%= "show-#{pet.id}" %>>
+      <%= form_with url: "/pet_applications?app_id=#{@application.id}", method: :post, local: true do |form| %>
```
juliet-e, line 40: Awesome RESTful route again! Another RESTful option if you want to avoid the query param would be:
"/applications/:application_id/pets" --> it still mentions both resources involved in the join but has a path parameter. Both this route and the one you used that references the joins table are great RESTful options!

config/routes.rb
```diff
@@ -36,4 +42,12 @@
...
   get "/veterinary_offices/:veterinary_office_id/veterinarians/new", to: "veterinarians#new"
   post "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinarians#create"
+
+  get "/applications/new", to: "applications#new"
+  get "/applications/:id", to: "applications#show"
+  post "/applications", to: "applications#create"
+  patch "/applications/:id", to: "applications#update"
+
+  post "/pet_applications", to: "pet_applications#create"
+  patch "/pet_applications/:id", to: "pet_applications#update"
```
juliet-e, line 52: Excellent RESTful routing -- nice work!

spec/features/admin/applications/show_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+    it "doesn't approve the application if one of the pets is rejected" do
+      visit "/admin/applications/#{@app_1.id}" 
+
+      first_pet = @app_1.pets.first
+        within "#approve-#{first_pet.id}" do
+          click_button("Approve")
+        end
+
+      last_pet = @app_1.pets.last
+          within "#approve-#{last_pet.id}" do
```
juliet-e, line 164: Nice use of within blocks to narrow scope of the page

spec/features/applications/show_spec.rb
```diff
@@ -0,0 +1,177 @@
...
+      click_button "Search"
+
+      expect(current_path).to eq("/applications/#{@app_1.id}")
+      within "#show-#{@pet_4.id}" do 
+        expect(page).to have_content("Toaster")
+        expect(page).to have_button("Adopt this pet")
+      end
+    end
+
+    it "won't respond to non existent dog name" do
```
juliet-e, line 88: Love the name of this test ;) Nice sad path testing!

spec/features/applications/show_spec.rb
```diff
@@ -0,0 +1,177 @@
...
+      expect(page).to have_field(:pet_name)
+
+      fill_in(:pet_name, with: "Lob")
+      click_button "Search"
+
+      expect(Pet.search("Lob")).to eq([@pet_2])
+      expect(page).to have_content("Lobster")
+    end
+
+    it "is not case sensitive when searching for available pets" do
```
juliet-e, line 165: Back in week 1, when we talked about unit test vs integration test, I mentioned the idea of the test pyramid. This is something you'll likely explore later on in the job, but I think it's worth mentioning here. You're doing an awesome job testing this search functionality. There are a bunch of edge cases around non-existent names, partial matches, etc. This is a good example of a case where it can be kind of a pain test test all the edge cases at both the integration and the unit level. It's important to test at both levels, but generally, when following the test pyramid, it's okay to focus on the edge cases more at the unit level and then just verify expected behavior (aka happy path) at the integration level. That means that all the different edge cases for this method could live at the model spec level (i.e. test Pet.search with all these different cases) and then you could just test it with 1-2 happy paths at the integration level. You don't have to have 4-5 edge case tests at *both* levels (unless you really want to!)

juliet-e: https://martinfowler.com/articles/practical-test-pyramid.html

spec/models/shelter_spec.rb
```diff
@@ -67,5 +118,19 @@
...
+        expect(@shelter_2.get_name).to eq("RGV animal shelter")
+      end
+    end
+
+    describe ".get_address" do 
+      it "uses a raw SQL query to return the a shelter's address" do 
+        expect(@shelter_1.get_address).to eq("Aurora, CO")
+        expect(@shelter_2.get_address).to eq("Harlingen, TX")
+      end
+    end
```
juliet-e, line 134: Excellent unit testing!

PR #185, opened 2023-12-14
Quin Nordmark, Logan Finnegan

n0rdie:


app/views/applications/show.html.erb
```diff
@@ -0,0 +1,53 @@
...
+
+<section class='submit-application'>
+  <% if @app.pets.count > 0 %>
+    <%= form_with url: "/applications/#{@app.id}/update", method: :patch do |f| %>
+      <%= f.label :reason_to_adopt, "Why I would make a good owner for these pet(s)?:" %>
+      <p><%= f.text_area :reason_to_adopt %></p>
+      <%= f.submit "Submit Application" %>
+    <% end %>
+  <% end %>
+</section>
```
abdulredd, line 53: As I was using the app in the browser I noticed 2 bugs. Your application allows for the same pet to be added to the same application more than once, and secondly, once the app is submitted the "Submit Application" button is stll there.

spec/features/admin/applications/show_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+        expect(page).to_not have_content("Application Rejected")
+        expect(page).to_not have_content("Application Approved")
+        # And instead I see buttons to approve or reject the pet for this specific application
+        expect(page).to have_button("Reject Application")
+        expect(page).to have_button("Approve Application")
+      end
+    end
+  end 
+end
+
```
abdulredd, line 92: Overall these are some great tests. I think the lines where you are reassigning `@app_3` because the status changed in the database, can be replaced with `@app_3.reload` which will do the same thing.

spec/features/admin/shelters/index_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      expect(@walrus_haven.name).to_not appear_before(@puppy_hope.name)
+      expect(@walrus_haven.name).to_not appear_before(@the_farm.name)
+      expect(@puppy_hope.name).to appear_before(@walrus_haven.name)
+      expect(@the_farm.name).to appear_before(@walrus_haven.name)
+      
+      expect(@puppy_hope.name).to_not appear_before(@the_farm.name)
+      
+      expect(@the_farm.name).to appear_before(@walrus_haven.name)
+      expect(@the_farm.name).to appear_before(@puppy_hope.name)
+    end
```
abdulredd, line 41: You should be fine with just having the `appear_before`, you do not the to check that something does NOT appear before something else, it's the same thing.

The only thing I would suggest further is to use `within` blocks to scope where you're looking for the specific content. This can make your tests more robust and less prone to false positives.

spec/features/applications/new_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+    # fill_in(:description, with: "")
+    # And I click submit
+    click_on("submit")
+    # Then I am taken back to the new applications page
+    # ???
+    # And I see a message that I must fill in those fields.
+    expect(page).to have_content("Application not created: Required information missing.")
+    expect(current_path).to eq("/applications/new")
+    end
+  end
```
abdulredd, line 35: I like this form test, but because we are checking for a form, I suggest making sure the form fields exit, even if you are not filling them out.

```ruby
# 3. Starting an Application, Form not Completed
it "returns an error message when form is not completed" do
  # When I visit the new application page
  visit "/applications/new"
  # Check that the form fields are still present
  expect(page).to have_field(:name)
  expect(page).to have_field(:street_address)
  expect(page).to have_field(:city)
  expect(page).to have_field(:state)
  expect(page).to have_field(:zip)
  expect(page).to have_field(:description)
  # And I fail to fill in any of the form fields
  # And I click submit
  click_on("submit")
  # Then I am taken back to the new applications page
  expect(current_path).to eq("/applications/new")
  # And I see a message that I must fill in those fields.
  expect(page).to have_content("Application not created: Required information missing.")
end
```

config/routes.rb
```diff
@@ -36,4 +36,16 @@
   get "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinary_offices#veterinarians"
   get "/veterinary_offices/:veterinary_office_id/veterinarians/new", to: "veterinarians#new"
   post "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinarians#create"
+
+  get "/applications/new", to: "applications#new"
+  post "/applications", to: "applications#create" 
+  get "/applications/:id", to: "applications#show"
+  patch "/applications/:id/update", to: "applications#update"
```
abdulredd, line 43: Should be: `patch "/applications/:id", to: "applications#update"`

config/routes.rb
```diff
@@ -36,4 +36,16 @@
...
+
+  post "/pet_apps", to: "pet_apps#create"
+  patch "/pet_apps/:id", to: "pet_apps#update"
```
abdulredd, line 46: For these join routes, I would suggest something like this:

```ruby
post "applications/:id/pets", to: "pet_apps#create"
patch "applications/:application_id/pets/:id", to: "pet_apps#update"
```

This is a more conventional approach, although what you have is not exactly, non-restful.

PR #186, opened 2023-12-14
Victor / Sam Paired Project

SamPuttman:


README.md
```diff
@@ -34,3 +34,8 @@ In this project, students will build upon the code in this repo to create a Pet
 
 This project will have one Check In. During your Check In, you should be prepared to review your database schema for the many to many relationship, your deployment to the internet, and your GitHub Projects board.
 
+## Database Design:
+
+https://dbdesigner.page.link/cxXa3wsL9DYfJx287
```
juliet-e, line 39: Thanks for including the db schema link!

app/controllers/admin/applications_controller.rb
```diff
@@ -0,0 +1,27 @@
+class Admin::ApplicationsController < ApplicationController
```
juliet-e, line 1: Nice controller organization here with the admin namespace!

app/controllers/admin/applications_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+  
+  def approve_pet
```
juliet-e, line 3: A more RESTful approach here would be to just have one `update` action in this controller and pass `approved` or `denied` as query parameters (will add a note in the view about how you can do this). These two actions are nearly identical so we don't want to duplicate logic, but make a more flexible update action. Also, per REST, we want to name our controller actions only using CRUD actions

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+  def new
+    @application = Application.new
+  end
+
+  def create
+    @application = Application.new(application_params)
+    if @application.save
+      redirect_to "/applications/#{@application.id}"   
+    else
+      render "new"
```
juliet-e, line 16: would've like to see flash messages here for descriptive feedback for the user

juliet-e: (US 3 covers this functionality)

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+    end
+  end
+
+  def show
+    @application = Application.find(params[:id])
+    @pets = @application.pets
+  end
+
+  def search_pets
```
juliet-e, line 25: Instead of making a separate action here, another option would be to pass the search term as a query parameter back to the same show page. Since you're rendering the same view here anyways, you could either turn this into a helper method or instead add the following block to your `show` action:
```
if params[:name].present?
    @search_results = @application.search_pets_by_name(params[:name])
end
```

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+    @application = Application.find(params[:id])
+    @search_results = @application.search_pets_by_name(params[:name])
+    render 'show'
+  end
+
+  def submit
```
juliet-e, line 31: Consider naming this action `update` since this is basically what type of action this is: an operation to update the status attribute of an application.

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,53 @@
...
+    @application = Application.find(params[:id])
+    if @application.update(application_params.merge(application_status: 'Pending'))
+      redirect_to "/applications/#{@application.id}"
+    else
+      render 'show'
+    end
+  end
+
+  def adopt_pet
+    application = Application.find(params[:id])
```
juliet-e, line 41: If we're thinking about CRUD actions, what's happening here is a `create`, as we are creating a joins record (PetApplication) for a given pet and application. However, we wouldn't want to put this create action in this controller because we already have one above. 

So, a better location for this code would be a joins controller - consider making a PetApplicationsController where you could have a create action that does this implementation: creates a join record for a given pet and app

app/models/application.rb
```diff
@@ -0,0 +1,25 @@
+class Application < ApplicationRecord
+  validates :applicant_name, presence: true
+  validates :address,        presence: true
+  validates :city,           presence: true
+  validates :state,          presence: true
+  validates :zip_code,       presence: true
+  validates :description,    presence: true
+
+  has_many :pet_applications
+  has_many :pets, through: :pet_applications
```
juliet-e, line 10: Nice many to many set up!

app/models/application.rb
```diff
@@ -0,0 +1,25 @@
...
+  
+  def search_pets_by_name(name)
+    Pet.search_by_name(name)
```
juliet-e, line 13: Check out the ApplicationRecord -- there is a search function already built for you there that you can use! It is inherited into all models because all models inherit from ApplicationRecord. That means, you could call `Pet.search(search_term)` without implementing search again. 

Also, this method doesn't include querying the application table at all so it would make more sense in the Pet model if it were not already available

app/models/pet.rb
```diff
@@ -10,4 +13,16 @@ def shelter_name
   def self.adoptable
     where(adoptable: true)
   end
+
+  def self.search_by_name(name)
+    where("name ILIKE ?", "%#{name}%")
```
juliet-e, line 18: Check out the ApplicationRecord -- there is a search function already built for you there that you can use! It is inherited into all models because all models inherit from ApplicationRecord. That means, you could call `Pet.search(search_term)` without implementing search again.

app/models/pet.rb
```diff
@@ -10,4 +13,16 @@ def shelter_name
...
+  end
+
+  def approved
```
juliet-e, line 21: Is the `approved` column on the pets table necessary?

app/models/shelter.rb
```diff
@@ -31,4 +31,13 @@ def alphabetical_pets
...
   end
+
+  def self.reverse_alpha_order
+    Shelter.find_by_sql("SELECT * FROM shelters ORDER BY name desc")
+    #shelters = Shelter.order(name: :desc)
+  end
+
+  def self.pending_applications
+    Shelter.joins(pets: :applications).where("applications.application_status = ?","Pending") 
+  end
```
juliet-e, line 42: Nice querying in these methods!

app/views/admin/applications/show.html.erb
```diff
@@ -0,0 +1,23 @@
...
+<p><strong>Zip code:</strong> <%= @application.zip_code %></p>
+<p><strong>Why I would make a good pet owner:</strong> <%= @application.description %></p>
+
+<p><strong>Applying for pets:</strong></p>
+
+<% @application.pet_applications.each do |pet_application| %>
+  <div id="pet_<%= pet_application.pet.id %>">
+    <p><%= pet_application.pet.name %></p>
+    <% if pet_application.status.nil? %>
+      <%= button_to "Approve", "/admin/applications/#{@application.id}/pets/#{pet_application.pet.id}/approve", method: :patch %>
```
juliet-e, line 15: Instead of having 2 different paths and 2 different controller actions for this, consider adding parameters to this button_to by adding `params: { approved: true }` or false if rejecting. Then this would just direct to an `update` controller action that could handle both cases by looking at `params[:approved]`

app/views/admin/shelters/index.html.erb
```diff
@@ -0,0 +1,13 @@
+<h1>Admin Shelters Index</h1>
+
+<% @shelters.each do |shelter|%>
+  <%= shelter.name%>
+<% end %>
+
+<h3>Shelters with Pending Applications</>
+<div id="pending_applications">
+  <% Shelter.pending_applications.each do |shelter|%>
```
juliet-e, line 9: We call model instance methods from views all the time, but when it comes to class methods, it's typically more conventional to call the class method from the controller. In the controller action, you could do: `@shelters_with_pending_apps = Shelter.pending_applications` and then iterate on that instance variable in the view.

app/views/applications/new.html.erb
```diff
@@ -0,0 +1,44 @@
+<h1>New Application</h1>
+
+<%= form_with url: "/applications", method: :post, html: { id: 'form' }, data: { turbo: false }, local: true do |form| %>
+
+  <% if @application.errors.any? %>
+    <div style="color: red;">
+      <p>Please fill in all required fields.</p>
```
juliet-e, line 7: Ahhh I see you've flashed errors in this way instead!

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,49 @@
...
+<p><strong>Zip code:</strong> <%= @application.zip_code %></p>
+<p><strong>Why my home is a good fit for a new pet:</strong> <%= @application.description %></p>
+
+
+<p><strong>Applying to adopt:</strong>
+<% @application.pets.each do |pet| %>
+  <%= link_to pet.name, "/pets/#{pet.id}" %>
+<% end %></p>
+
+<% if @application.pets.any? && @application.application_status != "Pending" %>
```
juliet-e, line 40: This can definitely be considering display logic so it's okay to include this in the view per MVC convention!

IMO it would also be okay to create a model method to encapsulate it if it were more complex, and name the method something like `submittable?` so that you can call it on the application here to determine if you should allow the submit. This is beneficial because then you would be able to unit test it at the model spec level. But in this case, I don't think this display logic is complex enough to warrant needing to do this

config/routes.rb
```diff
@@ -1,39 +1,53 @@
...
+  get     "/applications/:id",                       to: "applications#show"
+  post    "/applications",                           to: "applications#create"
+  get     "/applications/:id/search_pets",           to: "applications#search_pets"
+  patch   "/applications/:id/adopt_pet",             to: "applications#adopt_pet"
+  patch   "/applications/:id/submit",                to: "applications#submit"
+
+  get     "/admin/applications",                                      to: "admin/applications#index"
+  get     "/admin/applications/:id",                                  to: "admin/applications#show"
+  get     "/admin/shelters",                                          to: "admin/shelters#index"
+  patch   "/admin/applications/:application_id/pets/:pet_id/approve", to: "admin/applications#approve_pet"
```
juliet-e, line 15: See note in controller about consolidating this down into 1 action and passing query parameters to determine whether it's an approval or rejection: 
`patch "/admin/applications/:application_id/pets/:pet_id", to: "admin/applications_update"`

config/routes.rb
```diff
@@ -1,39 +1,53 @@
...
-  get "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinary_offices#veterinarians"
-  get "/veterinary_offices/:veterinary_office_id/veterinarians/new", to: "veterinarians#new"
-  post "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinarians#create"
+  get     "/", to: "application#welcome"
+
+  get     "/applications",                           to: "applications#index"
+  get     "/applications/new",                       to: "applications#new"
+  get     "/applications/:id",                       to: "applications#show"
+  post    "/applications",                           to: "applications#create"
+  get     "/applications/:id/search_pets",           to: "applications#search_pets"
```
juliet-e, line 8: The search term could be added as a query parameter here and could redirect back to show rather than having a separate controller action

config/routes.rb
```diff
@@ -1,39 +1,53 @@
...
+  patch   "/applications/:id/adopt_pet",             to: "applications#adopt_pet"
```
juliet-e, line 9: (See note in controller related to this comment): The more RESTful option for adding a pet to an application would be:
`post "/applications/:application_id/pets", to "application_pets#create"`

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,49 @@
+<h1> Application Details </h1>
+
+
+<% if @application.application_status != "Pending" %>
+  <p><strong>Add a Pet to this Application</strong></p>
+  <%= form_with url: "/applications/#{@application.id}/search_pets", method: :get do |form| %>
```
juliet-e, line 6: This form path could still be the `show` action but the search term would be passed as a parameter

db/migrate/20231210203557_add_approved_to_pets.rb
```diff
@@ -0,0 +1,5 @@
+class AddApprovedToPets < ActiveRecord::Migration[7.0]
+  def change
+    add_column :pets, :approved, :boolean
```
juliet-e, line 3: Was this column used?

spec/features/admin_applications/approve_pet_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    victor_app.pets << pet_1
+  
+    #When I visit an admin application show page ('/admin/applications/:id')
+    visit "/admin/applications/#{victor_app.id}"
+
+    #For every pet that the application is for, I see a button to approve the application for that specific pet
+    expect(page).to have_content("Approve")
+    
+    #When I click that button
+    within("div#pet_#{pet_1.id}") do
```
juliet-e, line 18: Great work using a within block to distinguish between buttons!

spec/features/applications/index_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    @pet_1 = @shelter_1.pets.create!(name: "Mr. Pirate", breed: "tuxedo shorthair", age: 5, adoptable: true)
+    @pet_2 = @shelter_1.pets.create!(name: "Clawdia", breed: "shorthair", age: 3, adoptable: true)
+    @pet_3 = @shelter_1.pets.create!(name: "Ann", breed: "ragdoll", age: 3, adoptable: false)
+    @victor_app.pets << [@pet_1, @pet_2]
+  end
+  it "display's the applicants info" do
+
+    visit "/applications"
+
+    expect(page).to have_content("Victor Antonio Sanchez")
```
juliet-e, line 16: This test could be more robust if it included multiple applications, to verify that the index page shows *all* the applications in the db, not just one

spec/models/pet_spec.rb
```diff
@@ -3,6 +3,8 @@
 RSpec.describe Pet, type: :model do
   describe "relationships" do
     it { should belong_to(:shelter) }
+    it { should have_many :pet_applications }
+    it { should have_many(:applications).through(:pet_applications) }
   end
```
juliet-e, line 9: Make sure you test every method you add to the model at the unit level! you're missing some model method tests here

PR #206, opened 2024-02-14
Odell and Martin: Adopt don't shop

odellmac4:
FINAL


app/controllers/admin/application_pets_controller.rb
```diff
@@ -0,0 +1,9 @@
+class Admin::ApplicationPetsController < ApplicationController
+
+  def update
```
juliet-e, line 3: Awesome job routing the create and update actions for adopting/approving pets to a joins controller!!

This is minor but for this controller action, even though it's triggered from an admin page, it would've been okay to put this update action in the other ApplicationPetsController rather than make a separate admin joins controller. Because there's no requirement  on the path for this update including `/admin`, there's only a requirement that the show page that it redirects back to includes `/admin`. This update action is sort of a `behind-the-scenes` action similar to the create action in the joins controller. It doesn't render a view that the user would see, but instead just redirects back to the admin dashboard. So, it's not strictly required that it be an `admin` namespaced controller. Let me know if this doesn't make sense and I'm happy to chat further.

app/controllers/application_pets_controller.rb
```diff
@@ -0,0 +1,12 @@
+class ApplicationPetsController < ApplicationController
+  # require 'pry'; binding.pry
+  def create
```
juliet-e, line 3: Beautiful RESTful implementation of this functionality!

app/models/pet.rb
```diff
@@ -10,4 +14,9 @@ def shelter_name
   def self.adoptable
     where(adoptable: true)
   end
+
+  def self.find_pet(pet)
+    Pet.where("name ILIKE ?", "%#{pet}%")
```
juliet-e, line 19: It looks like this isn't used for searching for a pet because you used the inherited `search` function (nice work!). So you could take this method out

app/models/shelter.rb
```diff
@@ -31,4 +33,23 @@ def alphabetical_pets
...
     adoptable_pets.where("age >= ?", age_filter)
   end
+
+  def self.reverse_order
+    find_by_sql("SELECT * FROM shelters ORDER BY shelters.name desc;")
+  end
+
+  def self.shelters_pending_applications
+   Shelter.joins(pets: :applications).where(applications: {status:"Pending"}).pluck(:name)
+  end
```
juliet-e, line 43: Great data querying in these model methods! I see you wrote this join in raw SQL below. Great practice! Did you get it working? I've only reviewed it briefly but it looks great to me

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,80 @@
...
+
+<% @application.pets.each do|pet|%>
+<%= link_to "#{pet.name}", "/pets/#{pet.id}" %>
+<%end%>
+
+
+
+
+<section class="pet">
+  <%= form_with url: "/applications/#{@application.id}", method: :get  do |f| %>
```
juliet-e, line 30: Did this form ever give you trouble with having `data: {turbo: false}`? It looks like it works on your deployment which is great, but I've seen this cause issues with Turbo before if it's not disabled. Just something to keep in mind! If you see weird issues with your form, add `data: {turbo: false}`

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,80 @@
...
+    <% end %>
+  <% end %>
+</section>
+
+<section class="why_I_would_make_a_good_owner">
+  <section class="nopets">
+ 
+    <% if @application.pets.present? %>
+      <% if @application.status == "In Progress" %>
+          <%= form_with(url: "/applications/#{@application.id}?why_i_would_make_a_good_owner=true", method: :patch ) do |f| %>
```
juliet-e, line 66: Would this functionality still work if you took out `why_i_would_make_a_good_owner=true` from the path? You're checking for the presence of that param in the controller, so just having a user fill out the form should add that parameter, and you shouldn't need it in the path as a query parameter as well

app/views/pets/_form.html.erb
```diff
@@ -0,0 +1,16 @@
+<%# locals: (path:, method:) - %>
+<%= form_with url: path, method: method, local: true do |f|%>
```
juliet-e, line 2: Great use of a partial!

config/routes.rb
```diff
@@ -36,4 +36,19 @@
   get "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinary_offices#veterinarians"
   get "/veterinary_offices/:veterinary_office_id/veterinarians/new", to: "veterinarians#new"
   post "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinarians#create"
+
```
juliet-e, line 39: Excellent RESTful routing!

db/seeds.rb
```diff
@@ -5,3 +5,20 @@
 #
 #   movies = Movie.create([{ name: "Star Wars" }, { name: "Lord of the Rings" }])
 #   Character.create(name: "Luke", movie: movies.first)
+aurora = Shelter.create!(name: "Aurora shelter", city: "Aurora, CO", foster_program: false, rank: 9)
```
juliet-e, line 8: It might be helpful for your deployment to add `destroy_all` commands at the top for every object, so that you can run `rails db:seed` on Heroku multiple times and not need to worry about duplicate data

example:
`Pet.destroy_all`
`Application.destroy_all`
...etc

spec/features/admin/applications/show_spec.rb
```diff
@@ -0,0 +1,126 @@
...
+        visit "/admin/applications/#{shaggy.id}"
+        # For every pet that the application is for, I see a button to approve the application for that specific pet
+        within "#Pet-#{pet_1.id}" do
+            expect(page).to have_button("Approve #{pet_1.name}")
+        end
+        within "#Pet-#{pet_2.id}" do
+          expect(page).to have_button("Approve #{pet_2.name}")
+        end
+
+        within "#Pet-#{pet_3.id}" do
```
juliet-e, line 27: Great use of within blocks

spec/features/applications/new_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+        fill_in :city, with: "LA"
+        fill_in :state, with: "Cali"
+        fill_in :zip_code, with: 7089
+        fill_in :description, with: "I'm here"
+
+        click_on("Submit")
+        
+        expect(current_path).to eq("/applications/new")
+
+        expect(page).to have_content('All fields must be filled in')
```
juliet-e, line 64: Great testing for sad paths! Keep in  mind that you can use `.errors.full_messages.to_sentence` to get specific errors from an object if you want to customize error messages for each case.

spec/models/shelter_spec.rb
```diff
@@ -67,5 +87,11 @@
         expect(@shelter_1.pet_count).to eq(3)
       end
     end
+
+    describe "#reverse_order" do
+      it 'returns shelter names in reverse alphabetical order' do
+        expect(Shelter.reverse_order).to eq([@shelter_2, @shelter_3, @shelter_1])
+      end
+    end
```
juliet-e, line 95: Great attention to both model and feature testing! Excellent work

app/views/applications/show.html.erb
odellmac4: we did not get any errors with this form when were working on it but I totally forgot to add the turbo statement so this is a good reminder!

db/seeds.rb
odellmac4: Got it! I would love to discuss this briefly with you live just to get a bit more clarification what that actually does/looks like in the application.

juliet-e: Absolutely! We can definitely take a look at this in your next project. Your seeds file there should also have destroy  commands first so as not to duplicate data

PR #207, opened 2024-02-15
Jess & Hoa

kohljd:
Project submission


app/controllers/admin/approvals_controller.rb
```diff
@@ -0,0 +1,8 @@
+class Admin::ApprovalsController < ApplicationController
+  def create
+    application_pet = ApplicationPet.find_by(application_id: params[:id], pet_id: params[:pet_id])
+    application_pet.approve
+
+    redirect_to "/admin/applications/#{application_pet.application.id}"
```
juliet-e, line 6: I appreciated your defense of this design and why you chose to have a separate controller and make it a POST. It is more RESTfully conventional to direct that button to "/applications/:app_id/pets/:pet_id" and route that to `application_pets#update` BUT your defense makes a lot of sense and there is definitely a tradeoff between RESTfulness and user-friendly pathing. I agree that adding `/approve` to the path can make it more clear what the operation will do. This is a great example of when RESTfulness can be broken if you have a good defense.

app/controllers/application_pets_controller.rb
```diff
@@ -0,0 +1,10 @@
+class ApplicationPetsController < ApplicationController
```
juliet-e, line 1: Great controller organization! Using a joins controller here and a `create` action is a very RESTful choice

You don't actually need the `new` action here, though, as that would be for a form for creating a joins record, and you are just triggering that creation from the applications show page

app/models/application.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  def can_reject?
+    application_pets.where(pet_status: "rejected").any?
+  end
+
+  def reject
+    self.rejected!
+  end
+end
```
juliet-e, line 39: Great encapsulation of this data logic into model methods, with very readable names

app/models/pet.rb
```diff
@@ -10,4 +13,12 @@ def shelter_name
   def self.adoptable
     where(adoptable: true)
   end
+
+  def self.search(name)
```
juliet-e, line 17: This is correct implementation, but one of the benefits of working in a brownfield project is getting to take advantage of some of the existing methods. In the `application_record.rb` file (the parent class to every  model), there is a `search` method defined, which is very similar to this! So, you can use that search function instead and you don't need one implemented in the Pet model

app/models/shelter.rb
```diff
@@ -4,6 +4,7 @@ class Shelter < ApplicationRecord
   validates :city, presence: true
 
   has_many :pets, dependent: :destroy
+  has_many :applications, through: :pets
```
juliet-e, line 7: Excellent work adding the additional association so that you can join shelters directly to applications

app/views/admin/shelters/index.html.erb
```diff
@@ -0,0 +1,18 @@
...
+<h2>Shelters with Pending Applications</h2>
+<div id="pending_apps">
+  <%@shelters_pending.each do |shelter| %>
+    <ul>
+      <li><%= shelter %></li>
+    </ul>
+  <% end %>
+</div>
+
+    let!(:) {}
```
juliet-e, line 14: I think this code might have been added accidentally

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,48 @@
...
+                <%= form.text_field :pet_name %>
+                <%= form.submit "Search" %>
+            <% end %><br>
+
+            <% if @pets %>
+                Search Result:
+                    <% @pets.each do |pet| %>
+                    <div id="pet_<%=pet.id%>">
+                      <li><%= link_to "#{pet.name}", "/pets/#{pet.id}"%> 
+                      <%= button_to "Adopt this Pet", "/applications/#{@application.id}/pets",  method: :post, params:{pet_key: pet.id}%>
```
juliet-e, line 31: Nice work finding the syntax to add a query parameter for the pet ID! Another option could be to add the pet ID to the path (`"/applications/:app_id/pets/:pet_id"`)
Either of these path options is RESTful. Adding the pet ID is not required just maybe slightly easier syntax

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,48 @@
...
+                      </li>
+                      <br>
+                     </div>
+                <%end%>
+            <%end%>
+        
+
+            <% if @application.pets.size > 0 %>
```
juliet-e, line 39: This is very minor and is the only MVC infraction I've seen (and the rubric allows for 1-2), but I think it would be best to encapsulate this into a model method, and you could name it something like `submittable?` or `ready_to_submit?` to increase readability of this line. This is technically a data calculation (could use .count to make it just one SQL query) so is best implemented in a model method.

config/routes.rb
```diff
@@ -36,4 +36,18 @@
...
+  get "/applications/new", to: "applications#new"
+  post "/applications", to: "applications#create", as: :applications
+
+  get "/applications/:id/pets/new", to: "application_pets#new"
+  post "/applications/:id/pets", to: "application_pets#create"
+  get "/applications/:id", to: "applications#show", as: :application
+  patch "/applications/:id", to: "applications#update"
+
+  get "/admin/shelters", to: "admin/shelters#index"
+  get "/admin/applications/:id", to: "admin/applications#show"
```
juliet-e, line 49: Great RESTful routing and for the 2 routes below, well done defending your choice when departing from RESTfulness! 💯

spec/features/admin/applications/show_spec.rb
```diff
@@ -0,0 +1,169 @@
...
+    describe "User Story 12" do
+      describe "When I visit an admin application show page" do
+        it "sees a button to approve each pet in the application, after I click on the button, I'm taken back to the admin application show page and see 'Approved' next to that pet" do
+          visit "/admin/applications/#{application_1.id}"
+
+          expect(page).to have_button("Approve", count: 2)
+          expect(page).to have_content(pet_1.name)
+          expect(page).to have_content(pet_2.name)
+
+          click_on "Approve", match: :first
```
juliet-e, line 27: Great use of `match` for differentiating between buttons. This is also where you could make use of within blocks if you needed to approve only one

spec/models/shelter_spec.rb
```diff
@@ -1,6 +1,23 @@
...
+  let!(:pet_6) {shelter_3.pets.create!(adoptable: true, age: 1, breed: "dalmatian", name: "Pongo II")}
+
+  let!(:application_1) {Application.create!(name: "Sally", street_address: "112 W 9th St.", city: "Kansas City", state: "MO", zip_code: "64105", description: "I love animals. Please let me have one.", status: "pending")}
+  let!(:application_2) {Application.create!(name: "Marcus", street_address: "100 Hennepin Ave.", city: "Minneapolis", state: "MN", zip_code: "55401", description: "Dogs are the best. Please let me have one.", status: "pending")}
+
+  let!(:application_pet_1) {ApplicationPet.create!(pet: pet_1, application: application_1)}
+  let!(:application_pet_2) {ApplicationPet.create!(pet: pet_5, application: application_2)}
+
   describe "relationships" do
     it { should have_many(:pets) }
```
juliet-e, line 22: Because you added the additional association to applications, remember to add that relationship test here:
`it { should have_many(:applications).through(:pets) }`

spec/models/veterinary_office_spec.rb
```diff
@@ -49,26 +47,26 @@
...
       end
     end
 
     describe '.office_vets_filtered_by_rating' do
       it 'filters the office vets based on given rating' do
-        expect(@vet_office_1.office_vets_filtered_by_rating(9)).to eq([@vet_1])
+        expect(vet_office_1.office_vets_filtered_by_rating(9)).to eq([vet_1])
       end
     end
   end
```
juliet-e, line 72: Great attention to testing both at the model and feature level! Well done

PR #208, opened 2024-02-15
Selena and Laura's Adopt Don't Shop Paired Project

laurarvegav:


config/routes.rb
```diff
@@ -36,4 +37,16 @@
   get "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinary_offices#veterinarians"
   get "/veterinary_offices/:veterinary_office_id/veterinarians/new", to: "veterinarians#new"
   post "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinarians#create"
+
+  get "/applications", to: "applications#index"
+  get "/applications/new", to: "applications#new"
+  post "/applications", to: "applications#create"
+  get "/applications/:id", to: "applications#show"
+  post "/applications/:id", to: "application_pets#new"
```
abdulredd, line 45: This is the only route that isn't exactly restful. The `new` action is used for retrieving a form for creating a new record, so this should be a `GET` method.

To make this more restful, this should be: `get "/applications/:id/pets/new", to: "application_pets#new"`

Taking a look at your controller, it looks like you are creating a new record so your chosen verb is correct but the path and action aren't. Here is what the route should look like for your intended usage: `post "/applications/:id/pets", to: "application_pets#create"`

app/controllers/application_pets_controller.rb
```diff
@@ -0,0 +1,10 @@
+class ApplicationPetsController < ApplicationController
+  def new
+    ApplicationPet.create!(
+      pet_id: params[:pet_id], 
+      application_id: params[:id]
+      )
+      
+    redirect_to "/applications/#{params[:id]}?adopt=true"
+  end
```
abdulredd, line 9: Since we are creating a new record in the database, this method should be `create`.

As for the query parameter, I'm not exactly sure what you are doing with it since all you want to do is redirect to the applications show page.

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+  def index
+    @applications = Application.all
+  end
+
+  def show
+    @application = Application.find(params[:id])
+    @pets = Pet.search(params[:add_pet_name]) 
+    @pet_added = params[:adopt] == 'true'
+    @app_submitted = @application.status != "In Progress" 
+  end
```
abdulredd, line 11: Overall great job with the implementation of your `show` action, but there seems to be a slight issue starting at line 8. 

In your show view there is this code:

```erb
<% @pets.each do |pet|%>
  <div id="pet-<%= pet.id %>">
    <li><%= pet.name %></li>
    <%= button_to "Adopt this Pet", "/applications/#{@application.id}?pet_id=#{pet.id}", method: :post, local: true%>
  </div>
<% end %>
```

And when looking at that, I start to wonder where `@pets` are coming from if we haven't even searched for any pets yet since if we look inside the `params` hash there is no key called `:add_pet_name`. Essentially what's happening is that this is returning `nil` and therefore returning a list of all of the pets, which shouldn't be happening. If we had 100 pets, that means all 100 would be called on every time we went to the show page.

The search for a pet should be set within a condition to see if that `:add_pet_name` key first exists in the `params` hash.

```ruby
def show
  @application = Application.find(params[:id])

  if params[:add_pet_name]
    @pets = Pet.search(params[:add_pet_name])
  end

  @pet_added = params[:adopt] == 'true'
  @app_submitted = @application.status != "In Progress"
end
```

This would also mean that the view would need to be updated because on page load, `@pets` would be `nil` and you cannot call `each` on `nil` so that too would also need to be placed within a conditional.

```erb
<% if @pets %>
  <% @pets.each do |pet|%>
    <div id="pet-<%= pet.id %>">
      <li><%= pet.name %></li>
      <%= button_to "Adopt this Pet", "/applications/#{@application.id}?pet_id=#{pet.id}", method: :post, local: true%>
    </div>
  <% end %>
<% end %>
```

Now for lines 9 and 10, I think you asked about feedback on this in your video. I would have handled both of these in the view. In the `ApplicationPetsController#new` I would remove the query params and do this instead in the view:

```erb
<% unless @application.pet_count <= 0 || @application.status != "In Progress" %>
  <h4>Submit My Application</h4>
  <%= form_with url: "/applications/#{@application.id}", method: :patch, local: true do |f| %>
    
    <%= f.label :adopting_reason %>
    <%= f.text_field :adopting_reason %>

    <%= f.submit "Submit"%>
  <% end %>
<% end %>
```

By doing this I do not have to keep track of an arbitrary `true` value and just check if this application has pets or not. Line 9 and 10 can be removed from the controller and the code in the view is just a little easier to read.

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,49 @@
...
+        <%= f.submit "Submit"%>
+      <% end %>
+  </section>
+
+  <section class="found_pet">
+    <% @pets.each do |pet|%>
+      <div id="pet-<%= pet.id %>">
+        <li><%= pet.name %></li>
+        <%= button_to "Adopt this Pet", "/applications/#{@application.id}?pet_id=#{pet.id}", method: :post, local: true%>
+      </div>
```
abdulredd, line 33: For this button, this route should be: `"/applications/#{@application.id}/pets"`

And you can then tag the query params on the end 

```erb
<%= button_to "Adopt this Pet", "/applications/#{@application.id}/pets?pet_id=#{pet.id}", method: :post, local: true %>
```

Or them in using a params hash

```erb
<%= button_to "Adopt this Pet", "/applications/#{@application.id}/pets", method: :post, local: true, params: { pet_id: pet.id } %>
```

spec/features/applications/show_spec.rb
```diff
@@ -0,0 +1,220 @@
...
+
+      # And I see all the pets that I want to adopt
+      within '.pets_in_application' do
+        expect(page).to have_content("Rocky")
+        expect(page).to have_content("Lucille Bald")
+      end
+
+      # And I do not see a section to add more pets to this application
+      expect(page).to_not have_content("Add a Pet to this Application")
+    end
```
abdulredd, line 160: Your tests for this US are great. Great use of `within` blocks!

spec/features/applications/show_spec.rb
```diff
@@ -0,0 +1,220 @@
...
+        click_button("Submit")
+      end
+
+      within ".found_pet" do 
+        # Then I see any pet whose name PARTIALLY matches my search
+        expect(page).to have_content("fluffy")
+        expect(page).to have_content("fluff")
+        expect(page).to have_content("Mr. fluff")
+      end
+    end
```
abdulredd, line 198: Another great test suite for this US, I don't have any suggestions for your tests. I think they are great.

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,47 @@
+class ApplicationsController < ApplicationController
+  def index
+    @applications = Application.all
+  end
+
+  def show
+    @application = Application.find(params[:id])
+    @pets = Pet.search(params[:add_pet_name])
```
abdulredd, line 8: I just want to clarify one thing for the `search` method that you are calling on Pet since I didn't hear it mentioned in the video. This method does not come pre-packaged with AR and is actually a user-defined method that is defined in the ApplicationRecord model:

```ruby
class ApplicationRecord < ActiveRecord::Base
  primary_abstract_class
  
  def self.search(search_params)
    where("name ILIKE ?", "%#{search_params}%")
  end
end
```

Because all models inherit from ApplicationRecord, we can also then use it on Pet.

app/controllers/admin/applications_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+    end
+    
+    @application_pet = ApplicationPet.find_by(
+      application_id: params[:id], 
+      pet_id: params[:pet_id]
+      )
+    @result = @application_pet&.status_update
+    
+    render :show
+  end
```
abdulredd, line 23: Some high-level code in this fine, great job!

app/models/shelter.rb
```diff
@@ -31,4 +31,12 @@ def alphabetical_pets
...
     adoptable_pets.where("age >= ?", age_filter)
   end
+
+  def self.order_by_reverse_alpha
+    find_by_sql("SELECT * FROM shelters ORDER BY name DESC")
+  end
+
+  def self.find_pending_applications
+    joins(pets: { application_pets: :application }).where(applications: { status: 'Pending' }).distinct
+  end
```
abdulredd, line 41: This is a great query, I see that you needed to get to the applications table through application_pets and you made use of your pets association to do that. An easier way to do this would be to add another association to the model:

```erb
has_many :application_pets, through: :pets
has_many :applications, through: :application_pets
```

And with this, you can simplify your query:

```ruby
def self.find_pending_applications
  applications.where(applications: { status: 'Pending' }).distinct
end
```

app/views/admin/applications/show.html.erb
```diff
@@ -0,0 +1,20 @@
...
+<h3> <%= @application.name %>'s Application</h3>
+
+<section class="pets">
+  <% @application.pets.each do |pet|%>
+    <div id="pet-<%= pet.id %>">
+      <li><%= pet.name %></li>
+      <% if !@application_pet || @application_pet.pet_id != pet.id %>
+        <%= button_to "Approve Application", "/admin/applications/#{@application.id}?pet_id=#{pet.id}&approve=true", method: :patch, local: true%>
+        <%= button_to "Reject Application", "/admin/applications/#{@application.id}?pet_id=#{pet.id}&reject=true", method: :patch, local: true%>
+      <% end %>
```
abdulredd, line 11: These routes aren't restful if we're thinking about what operation we are performing. We intend to update the status on the join table so the routes should look something like this: `"/admin/applications/#{@application.id}/pets/#{pet.id}?your_key=your_value"`

PR #209, opened 2024-02-15
Yain's submission

yainporter:
I think the original plan was to go over and refactor the code that was commented on from last inning's submission and work on the Extensions, but I ended up refactoring almost everything... It was fairly difficult to only do certain parts of it. I still feel like I may have misunderstood some of the User Stories. Writing the code was the simple part, I spent a lot of time just trying to make sense of the process of the approval/rejections. I did end up getting a couple of call backs working, and one scope in.


app/controllers/application_pets_controller.rb
```diff
@@ -0,0 +1,24 @@
+class ApplicationPetsController < ApplicationController
```
juliet-e, line 1: excellent refactoring to use the joins controller

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+      :name,
+      :street_address,
+      :city,
+      :state,
+      :zipcode,
+      :description,
+      :status)
+  end
+
+  def create_filter
```
juliet-e, line 27: The content of these controller actions look good! I'd say none of these are so complex that they *need* a helper filter method, so it would be okay to just have them living in the CRUD actions above. Creating these filters is not wrong, but might just be a bit overkill given the controller is concise enough

app/controllers/pets_controller.rb
```diff
@@ -49,4 +33,32 @@ def destroy
...
+      redirect_to "/pets/#{pet.id}"
+    else
+      redirect_to "/pets/#{pet.id}/edit"
+      flash[:alert] = "Error: #{error_message(pet.errors)}"
+    end
+  end
+
+  def create_filter
+    pet = Pet.new(pet_params)
+    if pet.save
```
juliet-e, line 57: If you're interested in trying to get rid of the `if/else` condition for error handling in the controller action (or if that conditional complexity is what makes you want to create a filter), another option is trying to build a global error handling method in the ApplicationController. This is something we'll see more in Mod 3, but take a look at this article and you could consider trying to create a parent method that rescues from ActiveRecord::RecordInvalid errors! https://www.martin-brennan.com/rails-global-rescue-from-error-application-controller/

app/controllers/veterinary_offices_controller.rb
```diff
@@ -64,4 +37,46 @@ def destroy
   def vet_office_params
     params.permit(:name, :max_patient_capacity, :boarding_services)
   end
+
+  def veterinarians_filter
+    @veterinary_office = VeterinaryOffice.find(params[:veterinary_office_id])
+    if params[:sort] == "alphabetical"
```
juliet-e, line 43: Instead of abstracting all the content from the controller action into another method, which still keeps the complexity all together, you could consider breaking up this method like so (also notice that this isn't a RESTful controller design, and could be handled in a different nested controller)

```
def veterinarians
       @veterinary_office = VeterinaryOffice.find(params[:veterinary_office_id])
       @office_vets = handle_params(params, @vetinerary_office)
end

... private

def handle_params(params, vet_office)
    if params[:sort] == "alphabetical"
      return vet_office.alphabetical_vets
    elsif params[:review_rating]
      return vet_office.office_vets_filtered_by_rating(
        params[:review_rating]
      )
    else
      return vet_office.on_call_vets
    end
end
```

app/models/application_pet.rb
```diff
@@ -0,0 +1,41 @@
...
+  validates :application_approved, inclusion: [true, false]
+  validates :application_approved, exclusion: [nil]
+  validates :application_reviewed, inclusion: [true, false]
+  validates :application_reviewed, exclusion: [nil]
+
+  belongs_to :application
+  belongs_to :pet
+  has_many :applications, dependent: :destroy
+
+  after_update :update_pet
```
juliet-e, line 12: Great work getting lifecycle callbacks working!

app/models/pet.rb
```diff
@@ -2,12 +2,16 @@ class Pet < ApplicationRecord
   validates :name, presence: true
   validates :age, presence: true, numericality: true
   belongs_to :shelter
+  has_many :application_pets, dependent: :destroy
+  has_many :applications, through: :application_pets, dependent: :destroy, source: :application
+
+  scope :adoptable, -> { where(adoptable: true) }
```
juliet-e, line 8: Nice work!

app/models/shelter.rb
```diff
@@ -4,31 +4,77 @@ class Shelter < ApplicationRecord
...
+        .joins(:application_pets)
+        .where("application_pets.application_approved = ?", false)
+  end
+
+  def application_id_for_pet(pet_id)
+    self.applications
+        .joins(:pets)
+        .where("pets.id = ?", pet_id)
+        .pluck(:id)
+        .first
```
juliet-e, line 78: Excellent, concise queries!! Well done!

app/views/admin/applications/_list_of_pet_applications.html.erb
```diff
@@ -0,0 +1,36 @@
...
+                                                          button: "Reject Pet Application"
+        }%>  <br>
+      <% elsif !pet_application.application_reviewed %>
+        <%= render partial: "approve_or_reject", locals: {
+                                                          pet_application_id: pet_application.id,
+                                                          app_id: @application.id,
+                                                          app_approved: true,
+                                                          button: "Approve Pet Application"
+        }%>  <br>
+          <%= render partial: "approve_or_reject", locals: {
```
juliet-e, line 23: Love the use of partials here to break up this hefty view!

app/controllers/veterinary_offices_controller.rb
juliet-e: I prefer this refactoring rather than just having the call to the filter in the controller action because with this type of structure, you're not just abstracting *everything*, so it's easier to see what instance variables will be available in that view. It is encapsulating just the more complex piece (the if/elsif/else statement) but keeping the standard skeleton for the controller action.

app/controllers/pets_controller.rb
juliet-e: This is just an alternative and possible optimization that we haven't learned about at all yet, but it might be fun for you to check out!

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      redirect_to show_applications_path(application)
+    end
+  end
+
+  def show_filter
+    @application = Application.find(params[:id])
+    if params[:pet_name]
+      @pets = Pet.search(params[:pet_name])
+      if @pets == []
```
juliet-e, line 50: Re:video - You do need this if statement because this creates the branch for the case in which there is an error (i.e. the pet wasn't found). You only want to add the flash key to the flash object when there is an error case. 

Great bug fix with remembering `flash.now`!!

app/views/layouts/application.html.erb
```diff
@@ -9,19 +9,35 @@
...
-    </div>
-  <% end %>
-    <section>
-      <p><%= link_to "Pets", "/pets" %></p>
-      <p><%= link_to "Shelters", "/shelters" %></p>
-      <p><%= link_to "Veterinarians", "/veterinarians" %></p>
-      <p><%= link_to "Veterinary Offices", "/veterinary_offices" %></p>
-    </section>
-    <%= yield %>
+    <% if !params[:pet_name] %>
```
juliet-e, line 36: I may need to play around with your code more, and you know it closer than I do, but I'm tempted to think that you don't need this conditional to only render the flahs if there's no pet_name. You should only be adding a flash message when there is an error case (i.e. when the pet name isn't found), so I believe you should be able to render this flash layout no matter what, and the message will just show up on the page if there is a message available. 

You would have to take it out of your application show view though, as you'll only want to render it once on the page

spec/features/admin/applications/show_spec.rb
```diff
@@ -0,0 +1,163 @@
...
+        click_button("Reject")
+      end
+
+      expect(page.current_path).to eq("/admin/applications/#{@application_2.id}")
+      expect(page).to have_content("Application for #{@application_2.name} has been rejected")
+    end
+  end
+
+  describe "User Story 17 - Application Approval makes Pets not adoptable" do
+    it "can change the status of a pet from `adoptable` to `not adoptable` when ALL pets are approved" do
```
juliet-e, line 105: Nice refactor here to add iteration to make test very robust but very readable!

spec/models/pet_spec.rb
```diff
@@ -12,28 +12,47 @@
...
+
+    @application_pet_1 = ApplicationPet.create!(application_id: @application_1.id, pet_id: @dog.id)
+    @application_pet_2 = ApplicationPet.create!(application_id: @application_1.id, pet_id: @cat.id)
   end
 
   describe "class methods" do
-    describe "#search" do
-      it "returns partial matches" do
-        expect(Pet.search("Claw")).to eq([@pet_2])
+    describe ".adoptable" do
```
juliet-e, line 32: Great work remembering to add tests for the callbacks and scope functionality!

app/views/layouts/application.html.erb
yainporter: I placed an if statement here because I wanted to relocate the flash message! So when the parameter for pet[:name] is present, I want to render the one inside my application show view right next to the search bar, opposed to at the top of the page. Would there be a better way of achieving this type of functionality? I'm going to assume it's more FE related!

app/controllers/pets_controller.rb
yainporter: Thanks for this tip, Juliet!

app/controllers/veterinary_offices_controller.rb
yainporter: When you are saying this isn't a RESTful controller design, are you referring to the fact you are redirecting to vet_offices?

juliet-e: I'm mostly referring to the fact that the application routes to a non-CRUD action in the controller, which is not the most RESTful routing (I'd opt for a nested vets controller under vet_offices)

app/views/layouts/application.html.erb
juliet-e: Ahhh I see! For the different placement, this makes sense! There are definitely many different ways to do this with various FE tools, but now that I understand that the 2 flash messages are distinct, I understand the need to have the if statement here!

PR #210, opened 2024-02-15
Jack Sweeney & Dana Howell

JackCSweeney:


spec/features/applications/show_spec.rb
```diff
@@ -0,0 +1,231 @@
...
+
+    fill_in "search", with:"#{pet.name}"
+    click_on "Search"
+    click_on "Adopt this Pet" 
+    fill_in "owner_endorsement", with: "I will be the best owner to this pet"
+    click_on "Submit Application"
+
+    expect(page).to have_content(pet.name)
+    expect(page).not_to have_content("Add a Pet To This Application")
+  end
```
abdulredd, line 176: I love that you separated the 2 test suites into logical chunks even though you can have probably stuffed them all into one `it` block. Great test setup and asserts. Great use of `to have_content` and `not_to have content` matchers.

My only suggestion is to maybe limit the amount of duplicate code setup you have by tucking this away into a before each. If for whatever reason you needed to update the schema of the applications table on the database, you now have several places to make those changes in your test as opposed to possibly just one.

config/routes.rb
```diff
@@ -1,13 +1,25 @@
...
   get "/shelters/:id", to: "shelters#show"
   post "/shelters", to: "shelters#create"
   get "/shelters/:id/edit", to: "shelters#edit"
   patch "/shelters/:id", to: "shelters#update"
   delete "/shelters/:id", to: "shelters#destroy"
+  
+  get "/admin/shelters", to: "admin_shelters#index"
+  get "/admin/applications/:id", to: "admin_applications#show"
+
+  patch "/admin/applications/:application_id/application_pets/:id", to: "application_pets#update"
```
abdulredd, line 22: Great job with your routes and making them very restful. I noticed that you went with joins controllers for your admin routes (`admin_shelters#index`, `admin_applications#show`). Another way you could have done this could be to use nested controllers (`admin/shelters#index`, `admin/applications#show`)

There isn't anything wrong with your approach, however.

One thing I do want to call out is line 22 and the path you are using. I think it would be better to have a route like this: `"/admin/applications/:application_id/pets/:id"` and in your controller fine the application_pet record with a `find_by` like so: `ApplicationPet.find_by(application_id: params[:application_id], pet_id: params[:pet_id])`.

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+      render :new
+    end
+  end
+
+  def show
+    @application = Application.find(params[:id])
+    if params[:search].present?
+      @pets = Pet.search(params[:search])
+    end
+  end
```
abdulredd, line 20: This method is super clean and so darn readable. Great job tucking your `@pets` behind a conditional to avoid `nil` errors in your view.

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,35 @@
+<%= render "applicant_info" %>
+
+<h3>Pets on Application:</h3>
+
+<%= render "pet_info" %>
```
abdulredd, line 5: Great job using your partial here.

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,35 @@
...
+    <% @pets.each do |pet| %>
+      <h3><%= pet.name %></h3>
+      <p>Age: <%= pet.age %></p>
+      <p>Breed: <%= pet.breed %></p>
+      <%= button_to "Adopt this Pet", "/applications/#{@application.id}/pets/#{pet.id}", method: :post %>
+    <% end %>
+  <% end %>
+<% end %>
+
+</section>
```
abdulredd, line 35: I'm just going to say that I think this code is beautiful... I'm going through it with my fine tooth comb and I really don't see anything I can call out.

I think the only thing is that this code could use some logical spacing to make it more readable and that your `</section>` should come before the `<% end %>` that closes the `<% if %>` on line 16.

Great job using conditionals to control what shows up and what doesn't, great job with sticking to REST principles in all your routes and verbs!

db/schema.rb
```diff
@@ -10,10 +10,33 @@
...
+  create_table "applications", force: :cascade do |t|
+    t.string "name"
+    t.string "street_address"
+    t.string "endorsement"
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.string "city"
+    t.string "state"
+    t.string "zip_code"
+    t.string "owner_endorsement"
```
abdulredd, line 36: I love that you realized that you would need to create a new column on applications table and went ahead and made that update. Oftentimes, we'll start with a rough idea of what the database should look like and then need to make modifications later on, great job doing that here with the addition of this column.

app/views/applications/_pet_info.html.erb
```diff
@@ -0,0 +1,19 @@
+<% @application.find_pets.each do |pet| %>
+  <div id="<%= pet.id %>">
+    <% if pet.find_application_pet(@application.id).status == "Pending" && local_assigns[:admin] %>
+      <%= pet.name %>
+      <%= button_to "Approve Application", "/admin/applications/#{@application.id}/application_pets/#{pet.find_application_pet(@application.id).id}?status=approve", method: :patch %>
+      <%= button_to "Reject Application", "/admin/applications/#{@application.id}/application_pets/#{pet.find_application_pet(@application.id).id}?status=reject", method: :patch %>
```
abdulredd, line 6: I left a comment in your `routes.rb` file but I will mention it again, to make your button routes more restful, they should be using this path: `"/admin/applications/#{@application.id}/pets/#{pet.id}?status=approve"`.

This way you can avoid having to call `pet.find_application_pet(@application.id).id`.

One other suggestion is another way to pass the status back to the controller without having to use query params.

```erb
<%= button_to "Approve Application", "/admin/applications/#{@application.id}/pets/#{pet.id}", method: :patch, params: { status: approve } %>
```

app/controllers/application_pets_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+  def update
+    if params[:status] == "approve"
+      application_pet = ApplicationPet.find(params[:id])
+      application_pet.update(status: "Approved")
+    elsif params[:status] == "reject"
+      application_pet = ApplicationPet.find(params[:id])
+      application_pet.update(status: "Rejected")
+    end
+    redirect_to "/admin/applications/#{params[:application_id]}"
+  end
```
abdulredd, line 21: This method is excellent, my only suggestion would be to eliminate the repetition you have on lines 14 and 17 since in both branches you need to find the ApplicationPet record so you can do that outside of the if.

```ruby
def update
  application_pet = ApplicationPet.find(params[:id])

  if params[:status] == "approve"
    application_pet.update(status: "Approved")
  elsif params[:status] == "reject"
    application_pet.update(status: "Rejected")
  end

  redirect_to "/admin/applications/#{params[:application_id]}"
end
```

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def update
+    application = Application.find(params[:id])
+    application.update(application_params)
+    application.update(status: "Pending")
+    redirect_to "/applications/#{application.id}"
+  end
```
abdulredd, line 27: Instead of having two separate update operations which would both make separates calls to the database, you can reduce this to one call by first adding to the `application_params` and then updating in one call.

```ruby
def update
  application = Application.find(params[:id])
  application.update(application_params.merge({ status: "Pending" }))
  redirect_to "/applications/#{application.id}"
end
```

This would also work:

```ruby
def update
  application = Application.find(params[:id])
  application_params[:status] = "Pending"
  application.update(application_params)
  redirect_to "/applications/#{application.id}"
end
```

We're just trying to limit the number of times we hit the database.

spec/features/admin/applications/show_spec.rb
```diff
@@ -0,0 +1,121 @@
...
+
+    visit "/admin/applications/#{@application_2.id}"
+
+    within "##{@pet_1.id}" do
+      expect(page).to have_button("Approve Application")
+      expect(page).to have_button("Reject Application")
+      expect(page).not_to have_content("Application Rejected")
+    end
+  end
+end
```
abdulredd, line 121: This is an amazing test sweet, seriously. 🤩 

Everything is here... you have `within` blocks, `before :each`, `to have`, `not_to have`, logical separation of assertions. These tests are great. Don't forget that you can also tuck your `visit` in the `before :each` as well.

app/controllers/application_pets_controller.rb
```diff
@@ -0,0 +1,27 @@
+class ApplicationPetsController < ApplicationController
+  
+  def new
+  end
+
+  def create
+    @application_pet = ApplicationPet.new(application_pets_params)
+    @application_pet.save
+    redirect_to "/applications/#{@application_pet.application_id}"
+  end
```
abdulredd, line 10: New and create actions are great places to make use of flash messages! What would happen if this action if the application_pet record didn't save?

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def create
+    application = Application.new(application_params)
+    if application.save
+      redirect_to "/applications/#{application.id}"
+    else
+      flash.now[:notice] = "Error: #{error_message(application.errors)}"
+      render :new
+    end
+  end
```
abdulredd, line 13: Great job using flash messages here.

app/models/application.rb
```diff
@@ -0,0 +1,24 @@
...
+  validates :name, presence: true
+  validates :street_address, presence: true
+  validates :city, presence: true
+  validates :state, presence: true
+  validates :zip_code, presence: true
+  validates :endorsement, presence: true
+
+  def find_pets
+    @pets = Pet.joins(:applications).where("applications.id = #{self.id}")
+  end
```
abdulredd, line 14: Since this method is not passing anything to a view, there is no need for an instance variable.

Beyond that, if the intent of this method is to get all the pets on this application, you already have that info through your association on line 3.

app/models/pet.rb
```diff
@@ -2,12 +2,18 @@ class Pet < ApplicationRecord
...
     shelter.name
   end
 
   def self.adoptable
     where(adoptable: true)
   end
+
+  def find_application_pet(application_id)
+    ApplicationPet.where("application_pets.application_id = #{application_id} AND application_pets.pet_id = #{self.id}").first
+  end
```
abdulredd, line 18: Based on the name of this method it looks like you are intending to retrieve a pet's application. In your controller I made a suggestion as to how you can do this much easier using `find_by`.

app/views/admin_applications/show.html.erb
```diff
@@ -0,0 +1,2 @@
+<%= render "applications/applicant_info" %>
+<%= render "applications/pet_info", admin: true %>
```
abdulredd, line 2: I wanted to talk about your use of partials here really quickly. This is an excellent job of making use of partials across directories. In situations where you find yourself using partials across directories, it's more common to find these partials in a `views/shared/` folder. This way your `render`s would look something like this:

```erb
<%= render "shared/applicant_info" %>
<%= render "shared/pet_info", admin: true %>
```

PR #211, opened 2024-02-15
Igor Magalhaes & Mel Langhoff

IgorrMagalhaess:


app/controllers/admin/adoption_applications_controller.rb
```diff
@@ -0,0 +1,22 @@
+class Admin::AdoptionApplicationsController < ApplicationController
+
+    def show
```
juliet-e, line 3: The idea you mentioned in the video about this action is a great one -- it would be better to break up this functionality into 2 actions

1) the show page, which displays a button to update the status of the ApplicationPet by approving or rejecting a pet
2) an update action that can receive the request from this button, update the ApplicationPet and then redirect back to this admin show page

The question of where that update action can live is a good discussion, too. We wouldn't want it to be an update in the PetsController or ApplicationsController because that is typically reserved for updating attributes on the Pet or Application object. What we're doing here is updating the joins table record, so this is a great use case for a **joins controller**. You can name it ApplicationPetsController or AdoptionApplicationPetsController, and it can have an update action where you change the status and then redirect to this admin page.

app/controllers/admin/adoption_applications_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+        @adoption_app = AdoptionApplication.find(params[:id])
+        @app_pets = @adoption_app.pets
+        @pets = Pet.search(params[:search])
+
+        if params[:approved_pet].present?
+            pet_id = params[:approved_pet]
```
juliet-e, line 9: A possible refactor: Consider If you changed the structure of your params so that you passed "approve" or "reject" as the value for a key, and the key could be called params[:update_status] or something generic. Then, you could just check for the presence of that parameter, and sending the value of that parameter (either "approve" or "reject") to a more generic model method that could have the if statement to check whether it should change to approved or rejected. This is an option for pushing some of that logic into a model method instead of a controller. It's totally okay to have if statements in the controller, but this is a case where you could simplify the controller  a bit and just have one if statement instead of 2

app/controllers/admin/shelters_controller.rb
```diff
@@ -0,0 +1,10 @@
+class Admin::SheltersController < ApplicationController
+
+    def index
+        @admin_shelters = Shelter.reverse_alphabetical_order
+
+        @shelters_with_pending_apps = Shelter.sort_by_pending
```
juliet-e, line 6: Great work encapsulating the querying at the model level and calling those methods here! Very MVC compliant

app/controllers/adoption_application_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+      @adoption_app = AdoptionApplication.find(params[:id])
+      
+      @pets = Pet.search(params[:search])
+
+      if params[:add_pet].present?
+         pet_id = params[:add_pet]
+         @adoption_app.add_pet_to_app(pet_id)
+      end
+
+      if params[:ownership_description].present?
```
juliet-e, line 13: Since this functionality changes an attribute on the AdoptionApplication (updates the status), a more RESTful option would be to have that "Submit Application" button go to an update action in this controller. That action can then redirect back to this show page.

app/controllers/adoption_application_controller.rb
```diff
@@ -0,0 +1,47 @@
+class AdoptionApplicationController < ApplicationController
+
+   def show
+      @adoption_app = AdoptionApplication.find(params[:id])
+      
+      @pets = Pet.search(params[:search])
+
+      if params[:add_pet].present?
```
juliet-e, line 8: This show action is doing **a lot** of different things, and you could break it up into a couple different actions instead in order to better comply with the Single Responsibility Principle and REST. 

When a user clicks on "Add Pet", the operation that will lead to is the creation of a joins record in your adoption_application_pets table. Since this is a `creation` operation, it's best for this to go to a `create` action. But, you're not creating new application, so it shouldn't go to a `create` action in this controller. Instead, you can create the joins controller I mentioned above. You could have an `AdoptionApplicationPetsController` with a `create` action, which can handle the creation of this joins record, and then redirect back to this show page

app/controllers/adoption_application_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+   def new
+   end
+
+   def create      
+      adoption_app = AdoptionApplication.new(adoption_app_params)
+      if adoption_app.save 
+         redirect_to "/applications/#{adoption_app.id}"
+      else
+         
+         flash.now[:alert] = "Error: #{error_message(adoption_app.errors)}"
```
juliet-e, line 28: Nice work making use of the inherited `error_message` method!!

app/models/adoption_application.rb
```diff
@@ -0,0 +1,25 @@
...
+   end
+
+   # admin 
+   def change_app_status(new_status)
+      self.update(status: new_status)
+   end
+
+   def add_ownership_description_to_app(description)
+      self.update(ownership_description: description)
+   end
```
juliet-e, line 24: Model looks good! These model methods are nice and readable. It's okay, however, to use `update` and `find` in controllers, though, just for the future! There's nothing wrong with further encapsulating this logic in a method like you have done here as long as you test it in the model spec.

app/models/adoption_application_pet.rb
```diff
@@ -0,0 +1,21 @@
+class AdoptionApplicationPet < ApplicationRecord
+   belongs_to :adoption_application
```
juliet-e, line 2: Great many to many set up!

app/models/adoption_application_pet.rb
```diff
@@ -0,0 +1,21 @@
...
+
+   validates :adoption_application_id, presence: true
+   validates :pet_id, presence: true
+   
+   def change_app_pet_status(new_status)
+      self.update_column(:status, new_status)
+   end
+
+   def self.change_status_to_approved(application_id, pet_id)
+      adoption_app_pet = where({pet_id: pet_id,adoption_application_id: application_id}).first
```
juliet-e, line 13: .where is a good choice here, and another option is `.find_by` which will allow you to pass in both `application_id` and `pet_id` and it will return just one object so you wouldn't have to call `first`. Either .where or `.find_by` are fine to use!

app/models/shelter.rb
```diff
@@ -31,4 +33,14 @@ def alphabetical_pets
...
   end
+
+  # admin
+
+  def self.reverse_alphabetical_order
+    find_by_sql("SELECT * FROM shelters ORDER BY name DESC;")
+  end
+
+  def self.sort_by_pending
+    joins(pets: :adoption_applications).where(adoption_applications: { status: "Pending" }).distinct.pluck(:name)
```
juliet-e, line 44: Excellent data querying in these model methods!

app/views/admin/adoption_applications/show.html.erb
```diff
@@ -0,0 +1,39 @@
...
+<ul>
+<% @app_pets.each do |pet| %>
+    <div id="pet-<%= pet.id %>">
+    <li><%= link_to "#{pet.name}", "/pets/#{pet.id}" %></li>
+    <ul>
+        <li>Breed: <%= pet.breed %></li>
+        <li>Age: <%= pet.age %></li>
+        <li>Adoptable: <%= pet.adoptable %></li>
+    </ul>   
+    <% if AdoptionApplicationPet.where(pet_id: pet.id,adoption_application_id: @adoption_app.id).first.status == "Approved" %>
```
juliet-e, line 23: `.where` is an AR method so we don't want to call it in the view per MVC conventions. Instead, you could create a model method in the Pet model that grabs the corresponding AdoptionApplicationPet with a given app ID. 

Another option would be to, on line 15, instead of iterating over an application's pets, you could iterate over an application's application_pets (i.e. iterate over the join objects) and then you could get the pet information by doing `adoption_application_pet.pet.breed` since each joins record belongs to the pet. And your if statement would then be simplified to `if adoption_application_pet.status == "approved"`

app/views/pets/new.html.erb
```diff
@@ -1,16 +1,6 @@
...
-
-  <%= f.label :age %>
-  <%= f.number_field :age %>
-
-  <%= f.label :adoptable %>
-  <%= f.check_box :adoptable, value: true %>
-
-  <%= f.submit %>
-<% end %>
+<%= render partial: 'form', locals:{
```
juliet-e, line 2: Nice refactoring with partials!

config/routes.rb
```diff
@@ -36,4 +37,14 @@
...
   get "/veterinary_offices/:veterinary_office_id/veterinarians/new", to: "veterinarians#new"
   post "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinarians#create"
+
+  get "/applications/new", to: "adoption_application#new"
+  get "/applications/:id", to: "adoption_application#show"
+  post "/applications", to: "adoption_application#create"
+
+  get "/admin/shelters", to: "admin/shelters#index"
+
+  get "/admin/applications/:id", to: "admin/adoption_applications#show"
```
juliet-e, line 47: Based on my comments from your controllers, there could be a couple other routes:

`post "/applications/:id/pets", to: "adoption_application_pets#create"` For adding a pet to an application (creating a joins record)
`patch "/applications/:app_id/pets/:id", to: "adoption_application_pets#update"` For approving or rejecting a pet on an app
`patch "/applications/:id", to: "adoption_applications#update"` for submitting an application and changing the status

db/seeds.rb
```diff
@@ -1,7 +1,28 @@
-# This file should contain all the record creation needed to seed the database with its default values.
-# The data can then be loaded with the bin/rails db:seed command (or created alongside the database with db:setup).
-#
-# Examples:
-#
-#   movies = Movie.create([{ name: "Star Wars" }, { name: "Lord of the Rings" }])
-#   Character.create(name: "Luke", movie: movies.first)
+
```
juliet-e, line 1: To prevent duplicate records in your deployment, you can start your seeds file with `destroy_all` commands:

Pet.destroy_all
AdoptionApplication.destroy_all
etc...

That way, your seeds file won't create multiple. You can't run `db:drop` on Heroku because it's a protected command so this is an easier option!

spec/features/admin/applications/show_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+ 
+
+            visit "/admin/applications/#{application_1.id}"
+
+            within "#pet-#{pet_1.id}" do
+                expect(page).to have_content(pet_1.name)
+                expect(page).to have_button("Approve")  
+            end
+            
+            within "#pet-#{pet_2.id}" do
```
juliet-e, line 51: Nice use of within blocks!

spec/features/adoption_application/show_spec.rb
```diff
@@ -0,0 +1,260 @@
...
+      click_button "Adopt Scooby"
+      
+      fill_in "search", with: "a"
+
+      expect(page).to have_content(pet_1.name)
+   end
+
+   it "shows a session to why the user would be a good owner after they add the pet to the application" do 
+      
+      shelter = Shelter.create(name: "Mystery Building", city: "Irvine CA", foster_program: false, rank: 9)
```
juliet-e, line 182: Great feature testing! For these next several tests that have some common test data, consider putting it in a before :each block, even if it's just for the tests in a describe block rather than for all tests in this file

spec/models/shelter_spec.rb
```diff
@@ -67,5 +70,24 @@
...
+        shelter_1 = Shelter.create(name: "Mystery Building", city: "Irvine CA", foster_program: false, rank: 9)
+        shelter_2 = Shelter.create(name: "Blah Shelter", city: "City CA", foster_program: true, rank: 4)
+        pet_1 = Pet.create(name: "Scooby", age: 2, breed: "Great Dane", adoptable: true, shelter_id: shelter_1.id)
+        pet_2 = Pet.create(name: "Scrappy", age: 1, breed: "Pit Mix", adoptable: true, shelter_id: shelter_2.id)
+        application_1 = pet_1.adoption_applications.create!(name: "Mel", street_address: "23 Main St", city: "Denver", state: "CO", zip_code: 80303, description: "I have a fenced backyard and love dogs", status: "Pending")
+        application_2 = pet_2.adoption_applications.create!(name: "Amy", street_address: "32 Central St", city: "Denver", state: "CO", zip_code: 80305, description: "I make a lot of money", status: "In Progress")
+        
+        expect(Shelter.sort_by_pending).to eq(["Mystery Building"])
+      end
+    end
```
juliet-e, line 91: Great attention to both model and feature tests!

PR #213, opened 2024-02-15
Lance & Nico pair project

LJ9332:


app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,30 @@
...
+        
+    end
+
+    def create
+        @application = Application.new(application_params)
+        if @application.save
+            redirect_to "/applications/#{@application.id}"
+        else
+            redirect_to "/applications/new"
+            flash[:alert] = "Error: #{error_message(@application.errors)}"
```
juliet-e, line 21: Nice work making use of the inherited `error_message` method to format those errors! Great sad path handling

app/controllers/pet_applications_controller.rb
```diff
@@ -0,0 +1,16 @@
+class PetApplicationsController < ApplicationController
```
juliet-e, line 1: Making a joins controller is a great idea, but you don't necessarily need a `index` or `show` in it. This joins controller can be helpful for implementing the functionality of adding a pet to an application. When a user clicks the "Add Pet" button, it could route to a `create` action in this joins controller because what that button is triggering is the **creation** of a joins record! So, that's a typical use case for a joins controller

app/models/application.rb
```diff
@@ -0,0 +1,7 @@
+class Application < ApplicationRecord
+    validates_presence_of :name, :street_address, :city, :state, :zip_code, :description, :status
+    
+    has_many :pet_applications
```
juliet-e, line 4: Nice many to many set up

app/views/application/new.html.erb
```diff
@@ -0,0 +1,22 @@
+<h1>New Application</h1>
+<%= form_with url: "/applications", local: true do |f| %>
```
juliet-e, line 2: While the default form method is POST, it might be helpful to explicitly add that to your form definition just so you are remembering how this form will route. Just a suggestion, not at all required

app/views/application/show.html.erb
```diff
@@ -0,0 +1,29 @@
...
+  <%= f.label :search, "Search for pets by name" %>
+  <%= f.text_field :search %>
+  <%= f.submit "submit" %>
+<% end %>
+<% if @search_results.any? %>
+  <h3> Search Results: </h3>
+  <ul>
+    <% @search_results.each do |pet| %>
+      <li><%= "#{pet.name}" %></li>
+      <li><%= button_to 'Adopt this Pet', "/pets/#{pet.id}", method: :get %></li>
```
juliet-e, line 19: As I mentioned in the comment in the PetApplicationsController, the most RESTful design for this functionality would be to route this button to that joins controller, and specifically to a create action for that joins controller (so the button method could be POST instead). A more RESTful path for that would also include both pets and applications, since you're creating a relationship between these 2 existing entities, i.e. creating a joins record. Consider either of these paths, both of which are RESTful:
`post "/applications/:application_id/pets/:pet_id", to "pet_applications#create"`
`post "/pet_applications", to "pet_applications#create"`

The benefit of the first option is that you can pass IDs in the path

app/views/application/show.html.erb
```diff
@@ -0,0 +1,29 @@
+<h1><%= @application.name %></h1>
+
+<h3> Street Address: <%= @application.street_address %></h3>
+<h3> City: <%= @application.city %></h3>
+<h3> State: <%= @application.state %></h3>
+<h3> Zip Code: <%= @application.zip_code %></h3>
+<h3> Description why you should have a dog: <%= @application.description %></h3>
+<h3> Add a Pet to this Application </h3>
+<%= form_with url: "/applications/#{@application.id}", method: :get, local: true do |f| %>
```
juliet-e, line 9: IF your form is acting funny but passing in tests, make sure you add `data: { turbo: false}`

config/credentials.yml.enc
```diff
@@ -1 +1 @@
-a0HWBU0Q5ZDoV4+RdAK5UFa9t+2CeiAmUvZYnd7PcNlSVrMDXH3ltjLBM5m9NgsREZmoyG4giJe1U+I3pq+NohlVl5d0H9PpQ2AiUPDD9ykQNylDaU16A70+TcslpVMmHJL7G4mwy+lU186pBJ482TKj8RR9L9g1wIKjy1GtKR84sudtiNcJQ0FNWqZ6Ett17jYB+F9qsBAcBDNWelUvEP9P0yOPTDxhWOhG4MaVbT3ejVw39k9OqX1e1IXMn6ias2ui30qOcnj1ssxeLlCcqimoOu5Ve8Gg+1mcRdOjwiR68Sli6CxCdk2BNCo9wfavVjpnCyEpKr2Wb/zhYYXNTzycUH4elfZ9McBOi0l+9GOKOtqoQS75MxBVlIFh9eh7HmatxV8uaeXj56Z2gf7YGbNV9ijNYS0zvnUO--4E6Rham/5eCiEFFZ--0Djqff/c3KFXDsunPKJVSQ==
\ No newline at end of file
+M4iZD8NDL07GXbjXu+IiSgjdjTvYr7fNBHH5oisR9lWf9JbUf3uDqPJ1KK8wVvHW7PWsqIxvw5rGgl7doJ0ATEW94Sr7GSQbaxmL07AoGby8us97BafSgBXFAHx0er7Dq6MT8kh4P7MmTRx4BmLqzdUdbBfWguvS2RYpTl64yyzc8j1IoArIwh0pD+BwKtdXAQJ5wbQPM4ZsFr0ZZQrddLmE/mW1dY/oUmtbi5IW8D9+UaZTa5ty5LK3SIjER+07sLS8or89SgIXCnjPLhr1bOcu7aNsu32pfkuZGPNXnUqRUwGlbKjJ22HULzSwdwCa5vIC0di46djIqKjOnjp6DO+EZOBD1YY+US3fWHmuC0DyFulJI/DoD8QIsBkopbZp5/ngwYVQRznpm+ZZ+yAQAmX4vAIV--mkhbqaZAybTDudHJ--PErwpyZvIVis1+jzuk5lsA==
```
juliet-e, line 1: You can add a file like this to your `.gitignore` so that you don't share your credentials on GH

config/routes.rb
```diff
@@ -1,5 +1,9 @@
 Rails.application.routes.draw do
   get "/", to: "application#welcome"
+  get "/applications/new", to: 'applications#new'
+  get "/applications/:id", to: 'applications#show'
+  post "/applications", to: "applications#create"
+
```
juliet-e, line 6: Nice RESTful routing here, but per previous comment, you'd probably want to break out the adding-pet-to-app functionality with this additional route: 
`post "/applications/:application_id/pets/:pet_id", to "pet_applications#create"`

spec/features/applications/create_spec.rb
```diff
@@ -0,0 +1,71 @@
...
+        # And I fail to fill in any of the form fields
+        fill_in "Name", with: "Lance"
+        # And I click submit
+        #save_and_open_page
+        click_button "submit"
+        # Then I am taken back to the new applications page
+        expect(page).to have_current_path("/applications/new")
+        # And I see a message that I must fill in those fields.
+        
+        expect(page).to have_content("Error: Street address can't be blank, City can't be blank, State can't be blank, Zip code can't be blank, Description can't be blank")
```
juliet-e, line 67: Great inclusion of both happy and sad path tests

spec/models/application_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'rails_helper'
+
+RSpec.describe Application, type: :model do
+    describe "Relationships" do 
+        it {should have_many :pet_applications}
```
juliet-e, line 5: Make sure you add this relationship test as well
`it { should have_many(:pets).through(:pet_applications) } `

spec/models/pet_spec.rb
```diff
@@ -3,6 +3,7 @@
 RSpec.describe Pet, type: :model do
   describe "relationships" do
     it { should belong_to(:shelter) }
+    it { should have_many :pet_applications}
```
juliet-e, line 6: Make sure you add this relationship test as well
it { should have_many(:applications).through(:pet_applications) }

PR #214, opened 2024-02-15
Logan Finnegan, Joey Reyes 

JRIV-10:


spec/features/admin/applications/show_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+      # Then I'm taken back to the admin application show page
+      expect(current_path).to eq("/admin/applications/#{@app_2.id}")
+
+      within "#app-status-#{@pet_3.id}" do 
+        # And next to the pet that I approved, I do not see a button to approve this pet
+        expect(page).to_not have_content("approve")
+        # And instead I see an indicator next to the pet that they have been approved
+        expect(page).to have_content("Status: Accepted")
+      end
+    end
```
abdulredd, line 44: This is a great test for this US and covers all the expected behavior. Great job using `within` to narrow your search field. The only suggestion I would make would be to change line 40 to check that that the page no longer has the button. While the current assertion you have works just fine, I just like to keep things consistent since you were checking for the page to have the button previously, it makes sense to check that it no longer has the button and not just the text.

config/routes.rb
```diff
@@ -36,4 +37,19 @@
...
+
+  get '/applications/new', to: "applications#new"
+  get "/applications/:id", to: "applications#show"
+  patch '/applications/:id', to: 'applications#update'
+  post '/applications', to: "applications#create"
+
+  post '/pet_apps', to: 'pet_apps#create'
+  
+  patch '/admin/pet_apps/:id', to: 'admin/pet_apps#update'
+
```
abdulredd, line 49: Overall these are some great RESTful routes. I do want to talk about lines 46 and 48 a little bit. While these routes are technically "RESTful" they are uncommon, you won't often see a path with the joins table in the name like this. A route like this would commonly be named like `patch "/admin/applications/:application_id/pets/:id"` and `post "/admin/applications/:id/pets"`.

This keeps the routes more RESTful.

app/views/admin/applications/show.html.erb
```diff
@@ -0,0 +1,12 @@
...
+    <h3>Pet Name: <%= pet_app.pet.name %></h3>
+    <% if pet_app.pet_status == "Pending" %>
+      <%= button_to "approve", "/admin/pet_apps/#{pet_app.id}?approve=true", method: :patch %>
+      <%= button_to "reject", "/admin/pet_apps/#{pet_app.id}?reject=true", method: :patch %>
+    <% else %>
+      <p> Status: <%= pet_app.pet_status %></p>
+    <% end %>
+  </div>
+<% end %>
+
```
abdulredd, line 12: Everything on this page is great! I'd like to suggest an alternative for your buttons to make them more restful and avoid having to use query params.

```erb
<%= button_to "approve", "/admin/applications/#{@app.id}/pets/#{pet_app.id}", method: :patch, params: { approve: true } %>
<%= button_to "reject", "/admin/applications/#{@app.id}/pets/#{pet_app.id}", method: :patch, params: { approve: true } %>
```

Doing it this way keeps your routes restful and allows you to still pass in the key and value that you want to make use of in the controller. With this change, however, you would need to update the code in your controller, to find a PetApp via the 2 ID's that you are passing in.

```ruby
class Admin::PetAppsController < ApplicationController 
  def update
    pet_app = PetApp.find_by(application_id: params[:application_id], pet_id: params[:id])

    if params[:approve]
      pet_app.update!(pet_status: 1)
    else
      pet_app.update!(pet_status: 2)
    end

    redirect_to "/admin/applications/#{pet_app.application.id}"
  end
end
```

app/controllers/admin/pet_apps_controller.rb
```diff
@@ -0,0 +1,11 @@
...
+  def update
+    pet_app = PetApp.find(params[:id])
+    if params[:approve]
+      pet_app.update!(pet_status: 1)
+    else
+      pet_app.update!(pet_status: 2)
+    end
+    redirect_to "/admin/applications/#{pet_app.application.id}"
+  end
+end
```
abdulredd, line 11: I suggested what this controller would look like in a comment in the show view.

app/models/pet_app.rb
```diff
@@ -0,0 +1,6 @@
+class PetApp < ApplicationRecord
+  belongs_to :application
+  belongs_to :pet
+
+  enum pet_status: ['Pending', 'Accepted', 'Rejected']
+end
```
abdulredd, line 6: Excellent job of making use of enums!

spec/features/applications/new_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+      fill_in("name", with: "Jake")
+      # And I click submit
+      click_button("submit")
+      # Then I am taken back to the new applications page
+      expect(current_path).to eq("/applications/new")
+      # And I see a message that I must fill in those fields.
+      expect(page).to have_content("Street address can't be blank, City can't be blank, State can't be blank, Zip can't be blank, and Description can't be blank")
+    end
+  end
+end
```
abdulredd, line 24: Great test. You have an empty before each at the top of this file that could probably be removed. The only other thing is your lone describe there on line 4 which reads "As a visitor". If I were onboarding as a new developer to this codebase and this test were to fail, the output in the terminal would be: "New Page As a visitor return an incomplete form" which isn't very descriptive. 

I understand that there are different ways to set up your test suite and your describes, I would just make them more descriptive and meaningful. Updating line 11 to `it 'when I submit an incomplete form I am re-routed to the form page and see a warning message'` or something like that to make it more descriptive.

app/views/applications/new.html.erb
```diff
@@ -0,0 +1,21 @@
...
+  <%= f.text_field :state %>
+
+  <%= f.label :zip %>
+  <%= f.text_field :zip %>
+
+  <%= f.label :description %>
+  <%= f.text_field :description %>
+
+  <%= f.submit "submit"%>
+<% end %>
```
abdulredd, line 21: 👍

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+  def create
+    new_app = Application.new(strong_params)
+    if new_app.save
+      redirect_to "/applications/#{new_app.id}"
+    else 
+      # require 'pry'; binding.pry
+      flash[:notice] = new_app.errors.full_messages.to_sentence
+      redirect_to "/applications/new"
+    end 
+  end
```
abdulredd, line 21: Great job using your flash message and using `to_sentence` on your errors.

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,39 @@
+class ApplicationsController < ApplicationController
+  def show 
+    @app = Application.find(params[:id]) 
+    @pets = @app.pets
+    if params[:search]
+      @found_pets = Pet.search(params[:search])
+    end
+  end
```
abdulredd, line 8: I noticed in your video that you had an error occur when you tried to visit an applications show page after creating an app and this is because of line 4. This should be nested inside a conditional because a new application would not have any pets to start with and therefore `@pets` would contain the value `nil` and you cannot iterate over `nil`.

```ruby
def show 
  @app = Application.find(params[:id]) 
  app_pets = @app.pets

  if app_pets # something like this could work
    @pets = app_pets
  end

  if params[:search]
    @found_pets = Pet.search(params[:search])
  end
end 
```

You would also need a conditional in your view to only iterate over `@pets` if `@pets` was not `nil`.

```erb
<% if @pets %>
  # Do the work...
<% end %>
```

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+  private 
+
+  def strong_params
+    params.permit(:name,
+                  :street_address,
+                  :city, 
+                  :state,
+                  :zip,
+                  :description)
+  end
```
abdulredd, line 38: Instead of `strong_params` I would name this `application_params` to be more meaningful.

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,39 @@
...
+      flash[:notice] = new_app.errors.full_messages.to_sentence
+      redirect_to "/applications/new"
+    end 
+  end
+
+  def update
+    app = Application.find(params[:id]) 
+    app.update!(description: params[:description], status: 1)
+    redirect_to "/applications/#{app.id}"
+  end
```
abdulredd, line 27: I suggest not using `!` methods in the controller and instead using an if/else using .save like you did in your create method. If for whatever reason line 25 didn't update, your app would result in an error page instead of handling it gracefully using a flash message or some other means.

app/controllers/pet_apps_controller.rb
```diff
@@ -0,0 +1,6 @@
+class PetAppsController < ApplicationController 
+  def create
+    PetApp.create!(application_id: params[:app_id], pet_id: params[:pet_id])
+    redirect_to "/applications/#{params[:app_id]}"
+  end
+end
```
abdulredd, line 6: Now that you know about flash messages, refrain from using `!` methods in your controllers.

app/models/application.rb
```diff
@@ -0,0 +1,14 @@
+class Application < ApplicationRecord
+	validates :name, presence: true
+	validates :street_address, presence: true
+	validates :city, presence: true
+	validates :state, presence: true
+	validates :zip, presence: true
+	validates :description, presence: true
+
+	enum status: ['In Progress', 'Pending', 'Accepted', 'Rejected']
```
abdulredd, line 9: Great use of enums.

app/models/shelter.rb
```diff
@@ -16,6 +16,16 @@ def self.order_by_number_of_pets
...
 
+  def self.reverse_alpha
+    find_by_sql("select * from shelters order by name desc")
+  end
+
+  def self.only_pending_apps
+    joins(pets: { pet_apps: :application })
+    .where(applications: { status:  1 })
+    .distinct
+  end
```
abdulredd, line 27: This is a great query, I see that you needed to get to the applications table through application_pets and you made use of your pets association to do that. An easier way to do this would be to add another association to the model:

```erb
has_many :application_pets, through: :pets
has_many :applications, through: :application_pets
```

And with this, you can simplify your query:

```ruby
def self.only_pending_apps
  applications.where(applications: { status: 1 }).distinct
end
```

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,50 @@
...
+  </section>
+<% end %>  
+
+<section class='found-pets'>
+  <% if @found_pets !=nil %>
+    <h2> Found pets: </h2>
+    <% @found_pets.each do |pet| %>
+      <div id="pet-<%= pet.id %>">
+        <h4><%= pet.name %></h4>
+        <%=button_to "Adopt this Pet", "/pet_apps?pet_id=#{pet.id}&app_id=#{@app.id}", method: :post%>
```
abdulredd, line 46: When you find yourself passing query params like this to pass along IDs, it's a good indication that your path might be incorrect.

```ruby
<%=button_to "Adopt this Pet", "/applications/#{@app.id}/pets", method: :post, params: { pet_id: pet.id }%>
```

PR #229, opened 2024-04-04
Aurora, Lance and Jared

JaredMHobson:


spec/features/admin/shelters/index_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+  it "has a section for shelters with pending applications" do
+    visit "/admin/shelters"
+
+    within "#pending_shelters" do
+      expect(page).to have_content("Aurora shelter")
+      expect(page).to have_content("Fancy pets of Colorado")
+      expect(page).to_not have_content("RGV animal shelter")
+    end
+  end
+end
```
abdulredd, line 53: There isn't a lot additional to add to these tests. The setup is great, the tests are great, you were able to target a specific part of the page using `within` blocks, great use of the `orderly` gem, and using `to_not` assertions. Overall great job! This is what a great test suite looks like.

config/routes.rb
```diff
@@ -1,6 +1,18 @@
...
+  get "/applications/new", to: "applications#new"
+  get "/applications/:id", to: "applications#show"
+  patch "/applications/:id", to: "applications#update"
+  post "/applications", to: "applications#create"
+
+  post "/applications/:id/pets", to: "pet_applications#create"
+
+  get "/admin/shelters", to: "admin/shelters#index"
+  get "/admin/applications/:id", to: "admin/applications#show"
+  get "/admin/applications/:id/pets/edit", to: "admin/pet_applications#edit"
```
abdulredd, line 13: The path in this route is incorrect. If we are editing a pet on an application then we would need that pet's ID. This route should be: `get "/admin/applications/:application_id/pets/:id/edit", to: "admin/pet_applications#edit"`

config/routes.rb
```diff
@@ -1,6 +1,18 @@
...
+  patch "/admin/pet_applications/:id", to: "admin/pet_applications#update"
```
abdulredd, line 14: While this route isn't technically incorrect, it is more common to see it as a nested resource route like so: `patch "/admin/applications/:application_id/pets/:id", to: "admin/pet_applications#update"` and in the `update` action in the controller you would use a `find_by` and use the `:application_id` and pet `:id` to find the appropriate pat application to make updates to it.

I know it may seem like a roundabout way to do it, but again is more conventional and will make a little more sense when we dig into nested routing.

app/controllers/admin/shelters_controller.rb
```diff
@@ -0,0 +1,5 @@
+class Admin::SheltersController < ApplicationController
+  def index
+    @shelters = Shelter.all
+  end
+end
```
abdulredd, line 5: Perfect.

app/models/shelter.rb
```diff
@@ -31,4 +32,15 @@ def alphabetical_pets
   def shelter_pets_filtered_by_age(age_filter)
     adoptable_pets.where("age >= ?", age_filter)
   end
+
+  def self.sort_reverse_alphabetically
+    find_by_sql("SELECT shelters.* FROM shelters ORDER BY shelters.name DESC")
+  end
```
abdulredd, line 38: Great job with your pure SQL query.

app/models/shelter.rb
```diff
@@ -31,4 +32,15 @@ def alphabetical_pets
...
+
+  def self.pending_shelters
+    joins(:applications)
+    .select('applications.status', 'shelters.*')
+    .where("applications.status = ?", "Pending")
+    .distinct
+  end
```
abdulredd, line 45: I love the way you used the association to `:applications` that you created on the `Shelter` model to make this join easier. Because of this association, you are able to join shelters, applications, pets, and pet applications. 

The only thing I will say about this query is that you do not need to use the sanitization syntax since you are not accepting input from somewhere. This is also acceptable: `.where("applications.status = Pending")`. If the value for `"Pending"` were coming from the user in a variable then you would want to sanitize that. (ie. `.where("applications.status = ?", status)`)

app/views/admin/shelters/index.html.erb
```diff
@@ -0,0 +1,18 @@
...
+<section id="all_shelters">
+  <% @shelters.sort_reverse_alphabetically.each do |shelter| %>
+    <div class="shelter_info">
+      <h3><%= shelter.name %></h3>
+      <p><%= shelter.city %></p>
+      <p><%= shelter.rank %></p>
+      <p><%= shelter.foster_program %></p>
+    </div>
+  <% end %>
+</section>
```
abdulredd, line 18: No feedback here, great job! Any reason why you used `class` on line 11 instead of `id`? It's not wrong, just curious.

spec/features/admin/applications/show_spec.rb
```diff
@@ -0,0 +1,137 @@
...
+        expect(page).to have_content("Approved")
+      end
+
+      within "#pet_#{@pet_3.id}_buttons" do 
+        expect(page).to have_button("Approve")
+        expect(page).to have_button("Reject")
+        expect(page).to_not have_content("Approved")
+      end
+    end
+  end
```
abdulredd, line 102: Lines 70 and 99, on pet number 3, it should have both buttons, which you have identified, and then you go on to say that it shouldn't have "Rejected" on line 70 and "Approved" on line 99. I think this could have been made better by checking that both "Approved" AND "Rejected" don't show up in both spots.

```ruby
within "#pet_#{@pet_3.id}_buttons" do 
  expect(page).to have_button("Approve")
  expect(page).to have_button("Reject")
  expect(page).to_not have_content("Approved")
  expect(page).to_not have_content("Rejected")
end
```

I can see you're asserting, however. That approving or rejecting a pet doesn't affect other pets. I would like the test to also show that approving or rejecting one pets doesn't have the opposite on other pets as well. Like approving pet 2 doesn't automatically reject all of the other pets as well.

app/controllers/admin/applications_controller.rb
```diff
@@ -0,0 +1,6 @@
+class Admin::ApplicationsController < ApplicationController
+  def show
+    @application = Application.find(params[:id])
+    @pet_applications = @application.pet_applications
+  end
+end
```
abdulredd, line 6: 👍

app/views/admin/applications/show.html.erb
```diff
@@ -0,0 +1,28 @@
...
+
+      <%=form_with url:"/admin/pet_applications/#{pet_app.id}", method: :patch, data: { turbo: false }, local: true do |f| %>
+        <%= f.hidden_field :status, value: "Rejected" %>
+        <%= f.hidden_field :application_id, value: pet_app.application_id %>
+
+        <%= f.submit "Reject" %> 
+      <% end %>
+    <% end %>
+  </div>
+<% end %>
```
abdulredd, line 28: There's a bitch of HTML missing in this view. Any test that we have should be wrapped in some sort of HTML tag, whether that be a header, paragraph, etc is up to you. We don't want tags just handing around all willy-nilly.

I also see that you've used forms in place of `button_to`... any reason? While `button_to` does transform into a form on the page in the browser, we encourage you to use the proper HTML in the view code. Looking over this code I imagine it is because you want to a couple of variables in hidden fields. This goes back to what I mentioned in the `routes.rb` file about nested routing. This eliminates the need for passing the application ID in a hidden field.

Here are a couple of ways you could have done this:

Query params
```erb
<%= button_to "Approve", "/admin/applications/#{@application.id}/pets/#{pet_app.pet.id}?status=Approved", method: :patch %>
<%= button_to "Reject", "/admin/applications/#{@application.id}/pets/#{pet_app.pet.id}?status=Rejected", method: :patch %>
```

Adding to param hash directly
```erb
<%= button_to "Approve", "/admin/applications/#{@application.id}/pets/#{pet_app.pet.id}", params: { status: "Approved" }, method: :patch %>
<%= button_to "Reject", "/admin/applications/#{@application.id}/pets/#{pet_app.pet.id}", params: { status: "Rejected" }, method: :patch %>
```

app/controllers/admin/pet_applications_controller.rb
```diff
@@ -0,0 +1,9 @@
+class Admin::PetApplicationsController < ApplicationController
+  def update
+    pet_application = PetApplication.find(params[:id])
+
+    pet_application.update(status: "#{params[:status]}")
+
+    redirect_to "/admin/applications/#{params[:application_id]}"
+  end
+end
```
abdulredd, line 9: I talked about this in a couple of places already but once your route is set up properly, the code in this method action would look similar, only line 3 would need to change.

```ruby
pet_application = PetApplication.find_by(application_id: params[:application_id], pet_id: params[:id])
```

app/assets/stylesheets/application.css
```diff
@@ -1,15 +1,90 @@
...
+    float: right;
+  }
+
+  .pet_info {
+    margin-bottom: 100px;
+  }
+
+  .shelter_info {
+    margin-bottom: 100px;
+  }
```
abdulredd, line 90: Love that you dove into some CSS ... ❤️

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+
+  def create
+    application = Application.new(application_params)
+    if application.save
+      redirect_to "/applications/#{application.id}"
+    else
+      redirect_to "/applications/new"
+      flash[:alert] = "error:#{error_message(application.errors)}"
+    end
+  end
```
abdulredd, line 21: Great use of sand path conditional logic and flash messaging!

PR #235, opened 2024-05-18
Experiment may 18

LancePants97:
Changed before(:all) blocks to before(:each) blocks to fix failing tests.


LancePants97:
Merging this myself since it was a simple change.

PR #237, opened 2024-05-23
Adopt Don't Shop

noahdurbin:
- completed through user story 14
- Aurora, Noah, and Jan
- https://adopt-dont-shop-2403-25fab3e2a83a.herokuapp.com/
- https://github.com/users/AuroraTuring/projects/1


app/controllers/admin/applications_controller.rb
```diff
@@ -0,0 +1,15 @@
+class Admin::ApplicationsController < ApplicationController
```
KatBrandt, line 1: Good job with the namespacing and file organization here.

app/controllers/admin/applications_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+  def show
+    @application = Application.find(params[:id])
+    @pet_applications = PetApplication.by_application_id(@application.id)
```
KatBrandt, line 4: Remember we have our associations, so all we have to do is call `@application.pet_applications` and we will get back all the records, so then we only have to send the `@application` instance variable.

app/controllers/admin/applications_controller.rb
```diff
@@ -0,0 +1,15 @@
...
+    @pets = Pet.search(params[:pet_name]) if params[:pet_name]
+  end
+
+  def update
+    @application = Application.find(params[:id])
+    @pet_application = @application.pet_applications.find_by(params[:pet_id])
+    if @pet_application.update(status: params[:status])
+      redirect_to "/admin/applications/#{@application.id}"
+    end
```
KatBrandt, line 13: And then what happens if the update fails? What would render? Make sure to always handle the happy and sad paths.

app/views/admin/applications/show.html.erb
```diff
@@ -0,0 +1,22 @@
...
+<% end %>
+<br>Application Status: <%= @application.status %></p>
+
+<% if @application.pets.count > 0 %>
+  <br><h2>All Pets: </h2>
+  <% @application.pet_applications.each do |pet_app| %>
+    <%= pet_app.pet.name %>
+    <% if pet_app.status == "Pending" %>
+      <%= button_to "Reject", "/admin/applications/#{@application.id}?status=Rejected", method: :patch, params: {pet_application_id: pet_app.id} %>
+      <%= button_to "Approve", "/admin/applications/#{@application.id}?status=Approved", method: :patch, params: {pet_application_id: pet_app.id} %>
```
KatBrandt, line 17: Nice use of these button helpers.

app/views/admin/applications/show.html.erb
```diff
@@ -0,0 +1,22 @@
+<%= render partial: "user_view", locals: { application: @application } %>
```
KatBrandt, line 1: Nice use of a partial.

app/models/pet_application.rb
```diff
@@ -0,0 +1,10 @@
+class PetApplication < ApplicationRecord
+  belongs_to :application
+  belongs_to :pet
+  validates :pet_id, presence: true
+  validates :application_id, presence: true
+
+  def self.by_application_id(application_id)
+    where(:id == application_id)
```
KatBrandt, line 8: usually you will see this as `where(id: application_id)` I don't even think having it with the `==` will return the correct results.

KatBrandt: Actually, what you'd want here would be `where(application_id: application_id)` since you don't want the id of the PetApplication record, you want the id of the application which is a foreign key.

config/routes.rb
```diff
@@ -1,39 +1,51 @@
...
+  patch '/veterinary_offices/:id', to: 'veterinary_offices#update'
+  delete '/veterinary_offices/:id', to: 'veterinary_offices#destroy'
+
+  get '/veterinarians', to: 'veterinarians#index'
+  get '/veterinarians/:id', to: 'veterinarians#show'
+  get '/veterinarians/:id/edit', to: 'veterinarians#edit'
+  patch '/veterinarians/:id', to: 'veterinarians#update'
+  delete '/veterinarians/:id', to: 'veterinarians#destroy'
+
+  get '/shelters/:shelter_id/pets', to: 'shelters#pets'
```
KatBrandt, line 44: `'shelter#pets'` is not a restful action. You could still have this one route to `'shetlers#show'` to show the pets for a specific shelter or you could even make another controller for ShelterPets and then navigated to the show action there.

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,36 @@
+class ApplicationsController < ApplicationController
+  def show
+    @application = Application.find(params[:id])
+    @pets = Pet.search(params[:pet_name]) if params[:pet_name]
+  end
+
+  def new; end
+
+  def create
+    @application = Application.new({ name: params[:name], address: params[:address],
```
KatBrandt, line 10: We want to use our strong params here to allow/not allow certain attributes to be sent in.

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,36 @@
...
+    end
+  end
+
+  def update
+    @application = Application.find(params[:id])
+    if @application.update(application_params)
+      redirect_to "/applications/#{@application.id}"
+    else
+      redirect_to "/applications/#{@application.id}"
+      flash[:alert] = "Error: #{error_message(shelter.errors)}"
```
KatBrandt, line 27: Usually we need to create the flash message before we do the redirect. I would think this wouldn't show until the next network request then, is there a test for this to verify that you can see the flash message?

config/routes.rb
```diff
@@ -1,39 +1,51 @@
...
-  get "/shelters/:shelter_id/pets", to: "shelters#pets"
-  get "/shelters/:shelter_id/pets/new", to: "pets#new"
-  post "/shelters/:shelter_id/pets", to: "pets#create"
-
-  get "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinary_offices#veterinarians"
-  get "/veterinary_offices/:veterinary_office_id/veterinarians/new", to: "veterinarians#new"
-  post "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinarians#create"
+  get '/', to: 'application#welcome'
+
+  get '/admin/shelters', to: 'admin/shelters#index'
```
KatBrandt, line 4: This is actually the correcvt long hand way to declare these routes. The "real way" that we'll learn later, is just a shorthand for this.

app/controllers/admin/shelters_controller.rb
```diff
@@ -0,0 +1,6 @@
+class Admin::SheltersController < ApplicationController
+  def index
+    @shelters = Shelter.reverse_alphabetical_shelters
```
KatBrandt, line 3: nit picking, but I would name this variable something more specific like `@shelters_reverse_alpha` to indicate they are sorted. Just my opinion!

spec/features/admin/shelters/index_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+          expect(page).to have_content('Whimsy Whiskers')
+        end
+
+        within '#shelter-1' do
+          expect(page).to have_content('Sandy Meows')
+        end
+
+        within '#shelter-2' do
+          expect(page).to have_content('Meow Wows')
+        end
```
KatBrandt, line 30: Great job with the within blocks

app/controllers/admin/applications_controller.rb
KatBrandt: I do realize this was not asked for though, just something to think about.

KatBrandt (commented):
Overall, great job here. Good just of routes, button_to and link_to helpers, tests look good. Please review my feedback and let me know if you have any questions.

PR #238, opened 2024-05-23
Chee, Garrett

cheeleertr:
To see admin links, click admin in upper left. While on admin index page, with keyboard, 
press up, up, down, down, left, right, left, right, b, a, enter
A little easter egg from Garrett to secure admin access.


spec/features/admins/applications_show_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+        end
+        it "Can approve a pet on an application" do
+            visit "/admin/applications/#{@app.id}"
+
+            within "#pet-request" do
+                expect(page).to have_link("Approve")
+            end
+
+            within "#pet-request" do
+            click_link "Approve"
```
KatBrandt, line 32: could probably move line 32 up to under line 28, that way you're checking for the link, then clicking it in that first block, then after that action happens, we can run all the expect statements in the second block on line 31. Just a nit-pick

spec/features/admins/applications_show_spec.rb
```diff
@@ -0,0 +1,82 @@
...
+                expect(page).to have_current_path("/admin/applications/#{@app.id}")
+                expect(page).to have_content("Approved")
+                expect(page).to_not have_link("Approve")
```
KatBrandt, line 35: I like that you checked that this other button is explicitly _not_ present.

config/routes.rb
```diff
@@ -36,4 +36,15 @@
...
+  get "/applications", to: "applications#index"
+  get "/applications/new", to: "applications#new"
+  post "/applications", to: "applications#create"
+  get "/applications/:id", to: "applications#show"
+  patch "/applications/:id", to: "applications#update"
+
+  get "/admin/shelters", to: "admin_shelters#index"
+  get "/admin/applications/:id", to: "admin_applications#show"
+  patch "/admin/applications/:id", to: "admin_applications#update"
+  get "/admin/index", to: "admin#index"
```
KatBrandt, line 49: This last route doesn't exactly follow RESTful routing patterns. This is like an admin home page, right? You could create a controller called Admin::HomeController, and then route to the index page there. That's actually not too different from what you did, but usually we don't use the words "index" and "show" in the urls since those are not too clear.

app/views/admin_applications/show.html.erb
```diff
@@ -0,0 +1,31 @@
...
+                    <%= link_to "Reject", "/admin/applications/#{@application.id}?reject=#{pet.id}", data: {turbo_method: :patch} , class: "btn btn-danger m-1" %>
+                <% elsif application_pet.approved == true %>
+                    <p> Approved </p>
+                <% elsif application_pet.approved == false%>
+                    <p> Rejected </p>
+                <% end %>
+            </div>
+        <% end %>
+    </div>
+</div>
```
KatBrandt, line 31: Instead of putting this in a file called `admin_applications`, you'd want to just put a folder called `applications` in the `admins` folder in the views, to the file path would be `views/admin/applications/show.html.erb`. You already have the namespace of `admin` from the parent directory.

app/views/admin_applications/show.html.erb
```diff
@@ -0,0 +1,31 @@
...
+    </div>
+    <div id="pet-request" class="m-1">
+        <h2>Pets to adopt</h2>
+        <% @application.pets.each do |pet| %>
+            <div class="m-2">
+                <p>Name: <%= pet.name %></p>
+                <p>Breed: <%= pet.breed %></p>
+                <p>Age: <%= pet.age %></p>
+                <p>Adoptable: <%= pet.adoptable %></p>
+                <% application_pet = @application.application_pets.find_by(pet_id: pet.id) %>
```
KatBrandt, line 19: Could abstract this logic by writing a method in the Application Model to find this record. Makes it more clean and we don't want this search logic in the view.

app/controllers/admin_applications_controller.rb
```diff
@@ -0,0 +1,17 @@
+class AdminApplicationsController < ApplicationController
```
KatBrandt, line 1: Instead of naming this controller like this, you can namespace it by creating the file path like this: `app/controllers/admin/applications_controller.rb`, then on line 1 here, it would be `class Admin::ApplicationsController < ApplicationController` - that is how you usually see it.

app/controllers/admin_applications_controller.rb
```diff
@@ -0,0 +1,17 @@
...
+
+    def update
+        @application = Application.find(params[:id])
+        if params[:approve]
+            pet = Pet.find(params[:approve])
+            @application.approve_pet(pet)
+        elsif params[:reject]
+            pet = Pet.find(params[:reject])
+            @application.reject_pet(pet)
+        end
```
KatBrandt, line 14: I like using the query params for this flow control. You could potentially abstract more of this logic into the Application Model. I like that you created the approve_pet and reject_pet methods, I'm wondering though if you couldn't have made a method for all this flow control logic to make your controller even thinner.  Overall, looks good though!

app/models/application.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    validates :name, presence: true
+    validates :street_address, presence: true
+    
+    def add_pet(pet)
+        self.pets << pet
+    end
+
+    def approve_pet(pet)
+        application_pet = ApplicationPet.find_by(application_id: self.id, pet_id: pet.id)
```
KatBrandt, line 13: Instead of calling the ApplicationPet table here, another option is to use the associations from Application, so this query would be `application_pet = application_pets.where(pet_id: pet_id)`. Since you already have that association setup, you've already done half the work!

app/models/application_pet.rb
```diff
@@ -0,0 +1,8 @@
+class ApplicationPet < ApplicationRecord
+    belongs_to :application
+    belongs_to :pet
+
+    validates :application_id, presence: true
+    validates :pet_id, presence: true
+    validates :approved, inclusion: { in: [true, false] }, allow_nil: true
```
KatBrandt, line 7: Nice job here with these restrictions and allowances.

app/models/shelter.rb
```diff
@@ -4,18 +4,29 @@ class Shelter < ApplicationRecord
   validates :city, presence: true
 
   has_many :pets, dependent: :destroy
+  has_many :applications, through: :pets
```
KatBrandt, line 7: Nice extension here

spec/features/admins/shelter_index_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+    it 'see a section with shelters listed reverse alphabetical by name' do
+      visit "/admin/shelters"
+      expect(page).to have_content("#{@shelter_1.name}")
+      expect(page).to have_content("#{@shelter_2.name}")
+      expect(page).to have_content("#{@shelter_3.name}")
+      expect("#{@shelter_2.name}").to appear_before("#{@shelter_1.name}")
+      expect("#{@shelter_2.name}").to appear_before("#{@shelter_3.name}")
+      expect("#{@shelter_3.name}").to appear_before("#{@shelter_1.name}")
+  end
+    it 'see name of every shelter that has a pending application' do
```
KatBrandt, line 41: More white space in this spec, a line break after the end on line 30 and a line break after line 40. Helps the eyeballs not see walls of text. :)

spec/features/admins/shelter_index_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+      @app2.add_pet(@pet_1)
+      @app2.add_pet(@pet_2)
+      @app1.add_pet(@pet_3)
+
+      visit "/admin/shelters"
+      expect(page).to have_selector("#with_pending_app")
+      expect(page).to have_content("Shelters with Pending Applications")
+
+      within "#with_pending_app" do
```
KatBrandt, line 50: Really great use of within blocks.

app/models/shelter.rb
```diff
@@ -4,18 +4,29 @@ class Shelter < ApplicationRecord
...
 
   def self.order_by_recently_created
     order(created_at: :desc)
   end
 
+  def self.order_by_reverse_alpha_sql
+    find_by_sql("SELECT * FROM shelters ORDER BY shelters.name desc")
```
KatBrandt, line 14: Nice job using the sql for this one!

app/models/shelter.rb
```diff
@@ -4,18 +4,29 @@ class Shelter < ApplicationRecord
...
     select("shelters.*, count(pets.id) AS pets_count")
       .joins("LEFT OUTER JOIN pets ON pets.shelter_id = shelters.id")
       .group("shelters.id")
       .order("pets_count DESC")
   end
 
+  def self.with_pending_app
+    joins(:applications)
+      .select("shelters.*, applications.status")
+      .where("applications.status = ?", "Pending")
```
KatBrandt, line 27: Nice job with this query!

app/controllers/admin_shelters_controller.rb
```diff
@@ -0,0 +1,6 @@
+class AdminSheltersController < ApplicationsController
+  def index
+    @shelters = Shelter.order_by_reverse_alpha_sql
```
KatBrandt, line 3: I might name this variable something to communicate these are not just all the shelters, but the ordered ones so `@shelters_reverse_alpha` or something. Just a nitpick :)

KatBrandt (commented):
Overall, really great job here team! Please review my feedback and let me know if you have any questions. As far as your question from your video about conventions, I made note of that here and there in my feedback, but otherwise, your code looked conventional (ie. good).

app/views/admin_applications/show.html.erb
```diff
@@ -0,0 +1,31 @@
...
+                <%if application_pet.approved == nil%>
+                    <%= link_to "Approve", "/admin/applications/#{@application.id}?approve=#{pet.id}", data: {turbo_method: :patch} , class: "btn btn-success m-1" %>
+                    <%= link_to "Reject", "/admin/applications/#{@application.id}?reject=#{pet.id}", data: {turbo_method: :patch} , class: "btn btn-danger m-1" %>
```
KatBrandt, line 22: Usually for actions that are like a toggle, we can use a `button_to` helper. Play with it and see which you like better!

PR #240, opened 2024-05-23
Austin C, Austin K

austincarrjones:


spec/features/pet_applications/new_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+
+			visit "/pet_applications/new"
+			fill_in("name", with: "Lassie")
+			click_button("Submit")
+			# save_and_open_page
+			expect(current_path).to eq("/pet_applications/new")
+			expect(page).to have_content("You must fill in all fields")
+    end
+	end
+
```
abdulredd, line 74: Just taking a look at the entire file as a whole, I think all of these tests are great. Since you have a pretty similar set up in each test, it might have been worth it to include a `before :each` block.

config/routes.rb
```diff
@@ -36,4 +36,12 @@
   get "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinary_offices#veterinarians"
   get "/veterinary_offices/:veterinary_office_id/veterinarians/new", to: "veterinarians#new"
   post "/veterinary_offices/:veterinary_office_id/veterinarians", to: "veterinarians#create"
+
+  get "/pet_applications/new" , to: "pet_applications#new"
+  post "/pet_applications", to: "pet_applications#create"
+  get "/pet_applications/:id", to: "pet_applications#show"
+  patch "/pet_applications/:id", to: "pet_applications#update"
+
+  post "/pet_applications/:app_id/pets/:pet_id", to: "pet_application_pets#create"
```
abdulredd, line 45: I think I might have been the one to give you this last route. This route should actually be `"/pet_applications/:app_id/pets"` and you would get the ID in the params hash either from a hidden field or something similar.

All the other routes are restful.

app/controllers/pet_applications_controller.rb
```diff
@@ -0,0 +1,32 @@
+class PetApplicationsController < ApplicationController
+  def new
+  end
+
+  def create
+    @application = PetApplication.create(name: params[:name], street: params[:street], city: params[:city], state: params[:state], zip: params[:zip], description: params[:description])
```
abdulredd, line 6: Instead of listing out all the attributes for creating a model object, you want to make use of strong params.

app/controllers/pet_applications_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+    if @application.save
+      @application.status_change("In Progress")
```
abdulredd, line 8: For the status change, I think it is much easier to set a default value for this column in your migration, then doing it here in the controller.

app/views/pet_applications/new.html.erb
```diff
@@ -0,0 +1,17 @@
...
+<%= form.text_field :city %>
+<%= form.label :state %>
+<%= form.text_field :state %>
+<%= form.label :zip %>
+<%= form.text_field :zip %>
+<%= form.label :description %>
+<%= form.text_area :description %>
+
+<%= form.submit "Submit" %>
+<% end %>
```
abdulredd, line 17: Just some lines needing indentation, but everything else in this file is great!

app/models/pet_application.rb
```diff
@@ -0,0 +1,16 @@
...
+  validates :state, presence: true
+  validates :zip, presence: true
+  validates :description, presence: true
+
+  has_many :pet_application_pets
+  has_many :pets, through: :pet_application_pets
+
+  def status_change(status)
+    self.status = status
+    save
```
abdulredd, line 14: If you do the status change in the controller before you cal `@application.save` then you would not need to call `save` here.

app/controllers/pet_applications_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+      redirect_to "/pet_applications/#{@application.id}"
+    else
+      redirect_to "/pet_applications/new", notice: "You must fill in all fields"
```
abdulredd, line 11: Great explanation in your video as to how you landed on this line.

spec/features/pet_applications/show_spec.rb
```diff
@@ -0,0 +1,145 @@
...
+			expect(page).to have_content("Add a Pet to this Application")
+			expect(page).to have_button("Search")
+
+			fill_in :pet_search, with: "Scooby"
+    	click_button("Search")
+			
+			expect(current_path).to eq("/pet_applications/#{application.id}")
+			expect(page).to have_content("Scooby")
+			expect(page).to_not have_content("Lobster")
+		end
```
abdulredd, line 73: This is a really great test.  Nice setup and adding 2 pets and checking that the other pet doesn't show up.

app/controllers/pet_application_pets_controller.rb
```diff
@@ -0,0 +1,7 @@
+class PetApplicationPetsController < ApplicationController
+  def create
+    @petapplicationpet = PetApplicationPet.create(pet_application_id: params[:app_id], pet_id: params[:pet_id])
+    # binding.pry
+    redirect_to "/pet_applications/#{params[:app_id]}"
+  end
+end
```
abdulredd, line 7: 💯

app/views/pet_applications/show.html.erb
```diff
@@ -0,0 +1,37 @@
...
+  <% end %>
+</ul><br>
+
+<% if @pet_application.status == "In Progress" %>
+  <p><strong>Add a Pet to this Application:</strong></p>
+  <%= form_with url: "/pet_applications/#{@pet_application.id}", method: :get, data: { turbo: false } do |f| %>
+    <%= f.label "Pet Name" %>
+    <%= f.text_field :pet_search %>
+    <%= f.submit "Search" %><br>
+  <% end %><br>
```
abdulredd, line 20: Great thorough explanation of your code here, you really spoke to it very well and articulated every line with a superb sense of understanding. That is exactly what these projects are supposed to be about. It's definitely nice to get to the end, but it's much better to have a full understanding of what you just did, like you displayed in the video.

app/views/pet_applications/show.html.erb
```diff
@@ -0,0 +1,37 @@
...
+
+  <% if @pets != nil %>
+    <% @pets.each do |pet| %>
+      <%= pet.name %> 
+      <%= button_to "Adopt this Pet", "/pet_applications/#{@pet_application.id}/pets/#{pet.id}", method: :post %>
```
abdulredd, line 25: I just want to circle back to this button because I think this is the one we worked on together, but I could also be wrong. For this button, in order to make the route more restful, here is what it would look like.

```ruby
<%= button_to "Adopt this Pet", "/pet_applications/#{@pet_application.id}/pets", params: {pet_id: pet.id}, method: :post %>
```

What you'll notice is that we added a new key/value pair of `params: {pet_id: pet.id}`. This is the exact same `params` hash that we get access to in the controller, we are accessing it and adding to it directly when this button gets selected.

As backend developers, our goal is to get the data we need in order to do the things we need to do on the server, this is yet another way to populate the `params` hash.

The route in `routes.rb` would then be `"/pet_applications/:app_id/pets"`.

In the controller, our `params` hash would contain `{app_id: <some num>, pet_id: <some num>}` where `app_id` comes from the dynamic segment in the path, and `pet_id` comes from us adding it directly in the `button_to`.

Reach out if this doesn't make sense.

app/views/pet_applications/show.html.erb
```diff
@@ -0,0 +1,37 @@
...
+    <% end %>
+  <% end %><br>
+
+  <p><strong>Submit this Application:</strong></p>
+  <p>Why would you make a good owner for these pets?</p>
+  <%= form_with url: "/pet_applications/#{@pet_application.id}", method: :patch, data: { turbo: false } do |f| %>
+    <%= f.label "Reasons:" %>
+    <%= f.text_area :reasons %>
+    <%= f.submit "Submit Application" %>
+  <% end %>
```
abdulredd, line 35: This form is excellent. Perfectly RESTful.

abdulredd (commented):
I know you didn't get a far as you both would have liked, but I think the code you have here is excellent. Your explanations in your video tells me that you are starting to put the pieces together and a mental model is starting to form in your minds.

I'm sure you both are going to make significant contributions to your group project. Don't hesitate to reach out if something doesn't make complete sense... I'm willing to break this stuff down to 1's and 0's if necessary.

Keep up the great work.

PR #241, opened 2024-05-23
Clyde Autin & Lance Nelson 2403 M2 Pair Project

LancePants97:
Adopt, don't shop!


spec/features/applications/show_spec.rb
```diff
@@ -0,0 +1,249 @@
...
+		visit "/applications/#{@app1.id}"
+	
+		expect(page).to have_content("Add a Pet to this Application")
+		expect(page).to have_button("Submit")
+
+		fill_in :search, with: "Rex"
+		
+		click_on("Submit")
+
+		expect(page).to have_current_path("/applications/#{@app1.id}?search=Rex")
```
abdulredd, line 192: `?search=Rex` is not part of a path, this assertion could also be `expect(page).to have_current_path("/applications/#{@app1.id}")`

spec/features/applications/show_spec.rb
```diff
@@ -0,0 +1,249 @@
...
+	end
+
+	it ' lets you add a pet to the application' do
+		visit "/applications/#{@app1.id}"
+
+		fill_in :search, with: "Rex"
+		click_on("Submit")
+		
+		expect(page).to have_content("Rex")
+		expect(page).to have_button("Adopt this Pet")
```
abdulredd, line 203: I would have liked to see you use a `within` block here. This application has 2 pets on it, there is nothing in this test block that says whose button this is.

```ruby
within "#pet-#{@pet1.id}" do
  expect(page).to have_content("Rex")
  expect(page).to have_button("Adopt this Pet")
  click_on("Adopt this Pet")
end
```

spec/features/applications/show_spec.rb
```diff
@@ -0,0 +1,249 @@
...
+		
+		click_on("Adopt this Pet")
+		
+		expect(page).to have_current_path("/applications/#{@app1.id}")
+		expect(page).to have_content("Rex")
```
abdulredd, line 208: You want to make sure that the button doesn't show up for Rex.

```ruby
within "#pet-#{@pet1.id}" do
  expect(page).to have_content("Rex")
  expect(page).to_not have_button("Adopt this Pet")
end
```

spec/features/applications/show_spec.rb
```diff
@@ -0,0 +1,249 @@
...
+		expect(page).to have_content("Mittens")
+		expect(page).to have_content("Buddy")
+		expect(page).to_not have_content("Adopt this Pet")
+	end
+
+	it "won't save if no pets are on an Applications page" do
+		visit "/applications/#{@app3.id}"
+
+		expect(page).to_not have_button("Submit Application")
+	end
```
abdulredd, line 232: Your test block says that it won't save if there are no pets on the Application, but we aren't clicking anything so I'm not sure how we are testing a save or no save feature. A better description would have been, "submit button not visible if application has no pets", or something similar.

```ruby
visit "/applications/#{@app3.id}"

expect(@app3.pets).to be_empty
expect(page).to_not have_button("Submit Application")
```

config/routes.rb
```diff
@@ -14,6 +27,7 @@
   get "/pets/:id/edit", to: "pets#edit"
   patch "/pets/:id", to: "pets#update"
   delete "/pets/:id", to: "pets#destroy"
+  get "/search_pets", to: "pets#search"
 
   get "/veterinary_offices", to: "veterinary_offices#index"
   get "/veterinary_offices/new", to: "veterinary_offices#new"
```
abdulredd, line 33: A lot of these routes are very restful, but there are some that are breaking REST.

We mentioned a while back that there may be a time when you will need to use a custom action, but those situations wouldn't come up in Mod 2.  These 2 routes should actually be just 1 route going to the update action.

```ruby
patch '/admin/applications/:id/approve', to: 'admin/applications#approve_app'
patch '/admin/applications/:id/reject', to: 'admin/applications#reject_app'

# this should be:
patch '/admin/applications/:id', to: 'admin/applications#update' # And to pass an approve or reject should be passed in as params
```

These 2 routes are also not restful. The first one is correct being a post, but the path, controller, and action are not correct. To be honest, I'm not sure what the second one is doing at all, since you already have a POST to CREATE for applications already. What is this second one supposed to be doing exactly?

```ruby
post "/applications/:id/add_pet/:pet_id", to: "applications#add_pet"
post "/applications/:id", to: "applications#submit"

These should be:
post "/applications/:id/pets", to: "application_pets#create"
```

This route is also not restful.

```ruby
get "/search_pets", to: "pets#search"
```

It seems like there are still a lot of questions surrounding restful routes. We have our advanced routing class coming up and this would be a great place to get any questions you still have answered.

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,50 @@
...
+<p>Street: <%= @application.street_address %></p>
+<p>City: <%= @application.city %></p>
+<p>State: <%= @application.state %></p>
+<p>Zip Code: <%= @application.zip_code %></p>
+<p>Why you are a good applicant: <%= @application.description %></p>
+<p>Application status: <%= @application.status %></p>
+<p>Pets in application:</p>
+<ul>
+  <% @application.pets.each do |pet| %>
+    <section id=<%= "pet-#{pet.id}" %>>
```
abdulredd, line 12: You have an opened a `<section>` tag and didn't close it.

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,50 @@
...
+    <li><%= pet.name %></li>
+  <% end %>
+</ul>
+
+<% if @application.status == 'In Progress' %>
+  <h2>Add a Pet to this Application</h2>
+  <%= form_with url: "/search_pets", method: :get, local: true do %>
```
abdulredd, line 19: This route is supposed to be getting the pet that you search for, so the route should be `"/applications/#{@application.id}"` and in the controller `show` action you should check whether there is a `:search` key inside of params and then find the pet and return it. You should not be sending this to a custom controller action.

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,50 @@
...
+    <div>
+    <%= label_tag :search, "Search for a pet by name" %>
+    <%= text_field_tag :search, params[:search] %>
+    <%= hidden_field_tag :application_id, @application.id %>
+    <%= submit_tag "Submit" %>
+    <% end %>
```
abdulredd, line 25: When using `form_with` it is conventional to pass in a block variable and use that for the label, text field, etc.

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,50 @@
...
+
+</div>
+
+<% if @pets %>
+  <h3>Search Results</h3>
+  <ul>
+    <% @pets.each do |pet| %>
+      <li>
+        <%= pet.name %> - 
+        <%= button_to "Adopt this Pet", "/applications/#{@application.id}/add_pet/#{pet.id}", method: :post %>
```
abdulredd, line 36: The path of this button is not correct. Here is what you want: `<%= button_to "Adopt this Pet", "/applications/#{@application.id}/pets", params: { pet_id: pet.id }, method: :post %>`.

We're passing the pet ID directly into the params hash.

app/views/applications/show.html.erb
```diff
@@ -0,0 +1,50 @@
...
+      </li>
+    <% end %>
+  </ul>
+<% end %>
+
+<% if @application.status == 'In Progress' && @application.pets.count >= 1  %>
+  <%= form_with url: "/applications/#{@application.id}", method: :post, local:true do |form| %>
```
abdulredd, line 43: Since we are making a modification to the application with a reason to why we would make a good owner, this would be a PATCH, not a POST.

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,47 @@
+class ApplicationsController < ApplicationController
+  def show
+    @application = Application.find(params[:id])
+
+    if params[:search].present?
+      @pets = Pet.where('name ILIKE ?', "%#{params[:search]}%")
```
abdulredd, line 6: A couple of things here:

1. `"%#{params[:search]}%"` does not need string interpolation since you are not adding is to text.
2. Doing a `where` in the controller is an MVC infraction.
3. The `ApplicationRecord` model already has a `search` method defined for you to use.

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def new
+  end
+
+  def submit
+    @application = Application.find(params[:id])
+    @application.update({"status" => "Pending"})
+    redirect_to "/applications/#{@application.id}"
+  end
```
abdulredd, line 17: You should not need to create custom actions.

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+      redirect_to "/applications/new"
+    end
+  end
+
+  def add_pet
+    @application = Application.find(params[:id])
+    pet = Pet.find(params[:pet_id])
+    @application.pets << pet
+    redirect_to "/applications/#{@application.id}"
+  end
```
abdulredd, line 35: You should not need to create custom actions.

Adding a pet to the application should be happening in an ApplicationPetController. This is the reason why you find yourself having to create extra actions, you should be creating controllers instead and sticking to the standard restful actions.

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def application_params
+    params.permit(:applicant_name, 
+                  :street_address, 
+                  :city, 
+                  :state, 
+                  :zip_code, 
+                  :description, 
+                  :status)
+  end
```
abdulredd, line 45: Strong params should be hidden behind a `private` since you will not need it outside of this class.

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+
+  def submit
+    @application = Application.find(params[:id])
+    @application.update({"status" => "Pending"})
+    redirect_to "/applications/#{@application.id}"
+  end
+
+  def create
+    @application = Application.new(application_params)
+    @application.status = 'In Progress'
```
abdulredd, line 21: Instead of setting every application status to "In Progress" here, you can also set a default value in your migration for this column.

spec/features/applications/new_spec.rb
```diff
@@ -0,0 +1,171 @@
...
+	end
+  it "keeps me on the same page and gives me an error message if I try to submit without filling out the form" do        
+    visit "/applications/new"
+
+    click_button ("Submit")
+	
+    expect(page).to have_current_path("/applications/new")
+    expect(page).to have_content("All fields must be filled in.")
+  end
+end
```
abdulredd, line 171: This one was a pretty straight forward test. No feedback from me.

app/views/applications/new.html.erb
```diff
@@ -0,0 +1,34 @@
...
+  </div>
+  <div>
+    <%= form.label :description %>
+    <%= form.text_area :description %>
+  </div>
+
+  <div>
+    <%= form.submit "Submit" %>
+  </div>
+<% end %>
```
abdulredd, line 34: This file contains ChatGPT generated code.

Also, is it me or do you have 2 new application forms? One in `app/views/applications/new.html.erb` and another similar one in `app/views/application/new.html.erb`

app/views/application/new.html.erb
```diff
@@ -0,0 +1,10 @@
+<%= form_with url: '/applications', method: :post, local:true do |form| %>
+  <p>Name: <%= form.text_field :applicant_name %></p>
+  <p>Street Address: <%= form.text_field :street_address %></p>
+  <p>City: <%= form.text_field :city %></p>
+  <p>State: <%= form.text_field :state %></p>
+  <p>Zip Code: <%= form.text_field :zip_code %></p>
+  <p>Description of why I would make a good home: <%= form.text_area :description %></p>
+
+  <%= form.submit "Submit" %>
+<% end %>
```
abdulredd, line 10: This form is great, the only thing is not to forget to make use of `form.label` for all the form fields. The only thing wrong with this file is that this is not the page that actually gets rendered to the browser. It is the other one in the `app/views/applications/` folder.

Your controller is called `ApplicationsController`, therefore any views for that controller would be in a plural `applications/` folder, not `application/` singular.

---

To answer your question whether you can make use of the form is multiple views and reduce the writing of the form to a couple of line, the answer is yes, by turning it into a partial.

db/schema.rb
```diff
@@ -10,10 +10,32 @@
...
   enable_extension "plpgsql"
 
+  create_table "application_pets", force: :cascade do |t|
+    t.bigint "application_id", null: false
+    t.bigint "pet_id", null: false
+    t.datetime "created_at", null: false
+    t.datetime "updated_at", null: false
+    t.index ["application_id"], name: "index_application_pets_on_application_id"
+    t.index ["pet_id"], name: "index_application_pets_on_pet_id"
+  end
```
abdulredd, line 24: Your join table should have a `status` column, this is the only way to properly implement US 12, 13, and 14.

app/controllers/admin/applications_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+      @application.approve
+      redirect_to "/admin/applications/#{@application.id}"
+    end
+
+    def reject_app
+      @application = Application.find(params[:id])
+      @application.reject
+      redirect_to "/admin/applications/#{@application.id}"
+    end
+end
```
abdulredd, line 22: For US 12, 13, 14 we are not approving or rejecting the application, we are approving or rejecting the specific pet on that application.

app/controllers/applications_controller.rb
```diff
@@ -0,0 +1,47 @@
...
+    if @application.save
+      redirect_to "/applications/#{@application.id}"
+    else
+      flash[:error] = "All fields must be filled in."
```
abdulredd, line 25: To answer your question about whether `"All fields must be filled in."` should go in the view since it seems to be view logic. The quick answer to that is no, we are storing that message into the `flash` hash. We then have code in our view that will look in that `flash` hash to see if there is an `:error` key and display the error, that is the view logic... aka using logic to determine if something should show up on the page.

To answer the second part of that question, similar to how the `params` hash is part of our controller, so is the `flash` hash, therefore you do not need to extract it to the model. It's available for you to use to display messages to the User, there is no data logic involved.

app/models/shelter.rb
```diff
@@ -31,4 +32,14 @@ def alphabetical_pets
   def shelter_pets_filtered_by_age(age_filter)
     adoptable_pets.where("age >= ?", age_filter)
   end
+
+  def self.order_by_name_desc
+    order(name: :desc)
+  end
```
abdulredd, line 38: This query was supposed to be an SQL only :)

app/views/admin/shelters/index.html.erb
```diff
@@ -0,0 +1,21 @@
...
+  <div id="shelter-<%= shelter.id %>">
+    <h3><%= shelter.name %></h3>
+    <p>Created at: <%= shelter.created_at %></p>
+    <%= link_to "Update #{shelter.name}", "/shelters/#{shelter.id}/edit" %>
+    <%= link_to "Delete #{shelter.name}", "/shelters/#{shelter.id}", method: :delete %>
+  </div>
+<% end %>
+
+<h2>Shelters With Pending Applications</h2>
+<%= @shelters.pending_apps.uniq.pluck(:name) %>
```
abdulredd, line 21: Calling `uniq` on a collection is a very expensive operation. Imagine you have 1,000,000 records and there are duplicates, the way you have it now, your application will retrieve all 1,000,000 records and then work through all of them to filter out the duplicates. 

A better way would be to not return duplicates in the first place. Because a database is optimized to remove duplicate records with a DISTINCT clause, we want to perform this operation on the database, thus limiting the number of entries that we pull into our application.

app/models/shelter.rb
```diff
@@ -31,4 +32,14 @@ def alphabetical_pets
...
+
+  def self.pending_apps
+    joins(:applications)
+      .select("shelters.*, applications.id AS application_id, applications.status")
+      .where("applications.status = ?", "Pending")
+  end
```
abdulredd, line 44: Add `.distinct` to remove duplicates

```ruby
def self.pending_apps
  joins(:applications)
    .select("shelters.*, applications.id AS application_id, applications.status")
    .where("applications.status = ?", "Pending")
    .distinct
end
```

app/models/pet.rb
```diff
@@ -10,4 +12,8 @@ def shelter_name
   def self.adoptable
     where(adoptable: true)
   end
+
+  def self.search(search)
+    where('name ILIKE ? OR breed ILIKE ?', "%#{search}%", "%#{search}%")
+  end
```
abdulredd, line 18: You did not need to define your own `search` method, since there is one in `ApplicationRecord` that you inherit. I guess it doesn't hurt to have your own version that overrides that one, since yours is also checking against the breed.

abdulredd (commented):
Great job getting through this project, I know it was a doozy. I will say that there are quite a few issues going on here and still a lot of areas of opportunity and growth for the both of you. 

This project did not meet expectations in Feature Completeness and Rails, and also lost points for not making use of `within` blocks in your test (you had 1 usage).

